PR #6, opened 2026-03-20
Use LICENSE file verbatim

### LINKED ISSUE #5: godoc not showing cos License
Trying to view the docs but message isl

> Documentation not displayed due to license restrictions.

https://pkg.go.dev/github.com/carlos7ags/folio@v0.4.0/layout

ebati:
The LICENSE file did not match the verbatim Apache text (https://www.apache.org/licenses/LICENSE-2.0.txt), so pkg.go.dev (pkgsite) did not recognize it.

The LICENSE should be added to the repository as-is, without modifying the APPENDIX. The updated text (with the correct year and copyright owner) will be placed in the header. Since you only included an SPDX identifier I continued with that notation and added the missing header too.

The changes in this PR are essentially the result of the following commands:

```sh
# Download LICENSE file
curl -o LICENSE https://www.apache.org/licenses/LICENSE-2.0.txt
# Add license headers to Go files
go install github.com/google/addlicense@latest
addlicense -l apache -c "Carlos Munoz and the Folio Authors" -y=2026 -s=only **/*.go
```

Fixes #5

### DIFF
--- a/cmd/folio/main.go
+++ b/cmd/folio/main.go
@@ -1,3 +1,6 @@
+// Copyright 2026 Carlos Munoz and the Folio Authors
+// SPDX-License-Identifier: Apache-2.0
+
 // Command folio is a CLI tool for PDF operations.
 //
 // Usage:


pedromorgan:
Yep that works..

but u removed carlos@all added at end.. 
- which is correct as per guidelines https://opensource.org/license/Apache-2.0

Can u add an AUTHORS as well with carlos.BDFL atop ;-))) ?

carlos7ags (approved):

PR #39, opened 2026-03-23
Prevent SSRF attacks by letting the user interceptor, modify or block…

piotrxp:
… images requested by html content

## Description

Added a user configurable image interceptor function that lets a user filter or modify any image loaded by the provided html content. 
The idea is to prevent SSRF attacks, which would let the attacker execute requests from within your network.

## Related issue

No related issue, just a new feature I needed.

## Checklist

- [x] Code is formatted (`gofmt -s -w .`)
- [x] Tests pass (`make test`)
- [x] New functionality includes tests
- [x] No references to external PDF libraries (clean room)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ jobs:
         uses: actions/setup-go@v6
         with:
           go-version: '1.25'
+          cache: true
 
       - name: Install qpdf
         run: sudo apt-get update && sudo apt-get install -y qpdf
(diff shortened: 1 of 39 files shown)


enquora:
> I like this patch. It's simple and useful IMHO.

Simple and useful are, in this case, orthogonal to maintainable and safe.

The problem is not with the patch - it is with expanding the usage case from simple renderer of static documents to a fully-fledged html client application. LLM assistance or not, it is a fool's errand to compete with complete web browsers.

Which is what the current state of affairs implies.

pedromorgan:
@enquora am curious why the negative about this patch ?

the way I see it:

- a func to handle external - handy to have.. eg blocking  stuff IN CODE
- negatives = none.. All it is is a simple func to allow app developer to block
- otherwise anything is let through regardless
- and u agree that having and option to rewrite the "path" is a problem

Additionally, am am curious what the problem is? have u got an user case ?
where should it be blocked ??

PR #187, opened 2026-04-11
Add Devanagari example program

carlos7ags:
## Summary

Adds a small runnable example under `examples/devanagari/` that mirrors the existing `examples/rtl/` and `examples/cjk/` patterns. The example loads a Devanagari-capable system font and lays out four short paragraphs, each chosen to exercise a different phase of the Indic shaper:

- `नमस्ते दुनिया` — plain consonant + dependent vowel sequence.
- `क्षत्रिय` — the kṣa conjunct, driven through the akhn (akhand) ligature.
- `कितना` — i-vowel sign (U+093F), driven through phase-4 pre-base matra reorder.
- `कर्म` — ra + halant at the head of a cluster, driven through rphf and the phase-4 reph move.

The layout pipeline routes Devanagari words through `ShapeDevanagari` automatically via `shapeAndMeasureWord` in `layout/paragraph.go`, so the example uses the ordinary `NewParagraphEmbedded` API with no direct shaper calls. That keeps the whole `main.go` under 80 lines and makes it a realistic usage sample rather than a shaper unit test.

Also wires the example into `README.md` and `examples/README.md` alongside the other entries.

## Base branch

This PR is based on `feat/devanagari-shaper` (#186), not `main`, because the Devanagari shaper has not merged yet. The example depends on `ShapeDevanagari`, `Word.GIDs`, `MeasureGIDs`, `EncodeGIDs`, and the `shapeAndMeasureWord` dispatch — all introduced in #186. After #186 merges, change this PR's base to `main` and let GitHub rebase; no manual fixes should be needed.

## Font discovery

The example does not ship a font. `loadDevanagariFont()` searches platform-specific system paths (macOS Supplemental, Linux Noto directories, Windows Fonts) and returns nil if none is available. When nil, the example prints a short message and exits with status 0, matching the rtl example's behaviour on hosts without Arabic/Hebrew fonts. macOS ships three Devanagari TTCs by default (`Devanagari Sangam MN.ttc`, `ITFDevanagari.ttc`, `DevanagariMT.ttc`), so the example runs out of the box on macOS; Linux CI hosts typically need `fonts-noto-core` or similar for `NotoSansDevanagari-Regular.ttf`.

## Trade-offs

- No example test file. Existing example directories have no `_test.go`, so this one follows the precedent and relies on `go build ./examples/devanagari/` plus manual `go run` for verification.
- PDF output path is the working directory (`devanagari.pdf`), matching `examples/rtl` (`rtl.pdf`) and `examples/cjk` (`cjk.pdf`). Not written under `examples/devanagari/` so the directory stays source-only.
- Four paragraphs is the minimum that covers the four distinct shaper paths the reviewer would want to see exercised end-to-end. More samples would bloat the example without teaching anything new.

## Test plan

- [x] `gofmt -s -w .` clean
- [x] `go vet ./...` clean
- [x] `golangci-lint run ./...` clean (0 issues)
- [x] `go test ./...` all packages green
- [x] `go build ./examples/devanagari/` succeeds
- [x] `go run ./examples/devanagari` on macOS produces a non-empty `devanagari.pdf` (~9.8 KB)
- [x] `pdftotext devanagari.pdf -` recovers all four Devanagari strings intact (verifies the ToUnicode / ActualText path preserves the original codepoints even though the draw path emits shaped GIDs)

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,6 +486,7 @@ Each [`examples/`](examples/) subdirectory is a self-contained `go run` demo:
 | Example | What it shows |
 |---|---|
 | [`hello`](examples/hello/) | Minimal one-page PDF |
+| [`devanagari`](examples/devanagari/) | Devanagari (Hindi, Sanskrit, Marathi) text shaping |
 | [`fonts`](examples/fonts/) | Standard, custom, and Unicode fonts (CJK, Cyrillic) |
 | [`links`](examples/links/) | Hyperlinks, bookmarks, internal navigation |
 | [`forms`](examples/forms/) | Interactive AcroForm fields |
--- a/examples/README.md
+++ b/examples/README.md
@@ -13,6 +13,7 @@ go run ./examples/hello
 ```
 examples/
 ├── hello/          # minimal one-page PDF
+├── devanagari/     # Devanagari (Hindi, Sanskrit, Marathi) text shaping
 ├── fonts/          # standard, custom, and Unicode fonts (CJK, Cyrillic)
 ├── links/          # hyperlinks, bookmarks, internal navigation
 ├── forms/          # interactive AcroForm fields (text, checkbox, radio, dropdown)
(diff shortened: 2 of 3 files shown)


carlos7ags:
Superseding with examples/indic/ — script-family grouping (rtl / indic / cjk) instead of per-language. See follow-up PR.

PR #204, opened 2026-04-21
optimize: Flate recompression of stream payloads (#203)

### LINKED ISSUE #203: optimize: Flate recompression of stream payloads with size-regression guard
## Background

#202 landed the first lossless rewrite pass (orphan sweep) and a cross-cutting size-regression guard primitive (\`sizeRegressionGuard\` in \`document/writer_size_guard.go\`). The guard runs a candidate transform, compares output sizes, and falls back to the baseline if the candidate is not strictly smaller. It is currently unused.

This issue tracks the first integration: a Flate recompression pass over indirect stream payloads, gated by the guard so it cannot enlarge a file.

## Scope

Walk \`Writer.objects\`, find each \`*core.PdfStream\`, and for every eligible stream:

1. Decompress the existing payload according to the stream's filter chain.
2. Re-deflate at \`zlib.BestCompression\` (RFC 1950 / ISO 32000-1 §7.4.4).
3. Pass the candidate bytes through \`sizeRegressionGuard\` against the original payload. Commit only on shrink.
4. When committing, update \`/Filter\`, \`/DecodeParms\`, and \`/Length\` to reflect the rewritten payload.

The pass is opt-in via a new field on \`document.WriteOptions\`:

\`\`\`go
RecompressStreams bool
\`\`\`

Default off. Documented as lossless and safe to combine with \`UseXRefStream\`, \`UseObjectStreams\`, and \`OrphanSweep\`.

## Eligibility

Streams whose filter chain contains a payload that Flate cannot improve on must be skipped before invoking the guard, since invoking deflate over already-compressed bytes is wasted work and the guard would just revert. Skip when the leaf filter is one of:

- \`DCTDecode\` (§7.4.8) — JPEG payload
- \`JPXDecode\` (§7.4.9) — JPEG 2000 payload
- \`CCITTFaxDecode\` (§7.4.7) — bilevel CCITT G3/G4
- \`JBIG2Decode\` (§7.4.8) — bilevel JBIG2

Streams with an empty or absent \`/Filter\` chain are eligible. Streams with an existing \`FlateDecode\` filter are eligible — the original may have been compressed at a lower level by a non-Folio producer.

Where the win actually comes from today (Folio always writes \`BestCompression\` for its own streams):

- Imported PDFs round-tripped through \`reader/\`. \`reader/import.go:136\` writes inflated content streams without recompression, leaving an obvious win on the table.
- Imported streams whose source producer used a lower zlib level.
- Future passes that materialize uncompressed intermediate streams.

## Encryption interaction

\`/ObjStm\` and the standard security handler interact with payload bytes (§7.6.1). The recompression pass must run before the encryption walk so the encryption operates on the rewritten payload. Practically: refuse the pass with encryption enabled in the first cut, matching the precedents set by \`UseObjectStreams\` and \`OrphanSweep\`. Lift the restriction in a follow-up once the order-of-operations is exercised end-to-end.

## API surface

Additive on \`document.WriteOptions\`. Composes with the existing toggles:

\`\`\`go
type WriteOptions struct {
    // ... existing fields
    RecompressStreams bool
}
\`\`\`

Behavior contracts to preserve (consistent with #202):

- Zero-value \`WriteOptions\` produces byte-identical output to today.
- Refused on encrypted documents in the first cut.
- The guard skipping a stream is silent and individual; one stream regressing does not abort the whole pass.

## Acceptance

- [ ] \`WriteOptions.RecompressStreams\` toggle, default off, godoc with eligibility rules and the §7.4.4 reference.
- [ ] Pass implementation in a new \`document/writer_recompress.go\` (separate file so the dispatch ordering in \`writer_xref_stream.go\` reads as a list of toggles).
- [ ] Unit tests:
    - Recompresses a stream that was written at \`BestSpeed\` and shrinks it.
    - Skips DCT/JPX/CCITT/JBIG2 leaf-filter streams (no decode attempted).
    - Guard reverts a stream whose Flate output equals or exceeds the input (e.g., random bytes).
    - Encryption refusal: refuses up front, leaves writer state unmutated, retry without the option succeeds.
    - Composes with \`OrphanSweep\` (sweep first, then recompress on the kept set).
    - Composes with \`UseXRefStream\` and \`UseObjectStreams\`.
    - Idempotence: recompressing twice produces byte-identical output.
- [ ] End-to-end size-regression test using an imported fixture (round-trip a PDF through \`reader\` and \`document\`, assert the recompressed write is smaller than the un-recompressed write).
- [ ] \`examples/optimize\` extended with a fourth column or a separate fixture that demonstrates the win on a reader-imported document.

## Out of scope

- Recompression of stream payload that is itself an \`/ObjStm\` body — that interacts with the object stream packer's eligibility rules and is better handled inside that pass.
- Any decoder for filters Folio does not yet support (LZWDecode is decode-only; ASCII85/ASCIIHex are handled by the reader).
- Tuning zlib parameters per stream (window size, strategy). \`BestCompression\` is the single knob.

## Relation to #201

This is part of the \"Flate recompression with size-regression guard\" line item in #201. Splitting it out keeps each PR small and makes the guard's first integration testable in isolation.

carlos7ags:
## Summary

Closes #203. Adds the second integration of the optimizer's lossless rewrite passes and the first real caller of the size-regression guard primitive shipped in #202.

- **\`WriteOptions.RecompressStreams\`** — walks every registered indirect object, finds eligible \`PdfStream\` payloads, and re-Flate-compresses them at \`zlib.BestCompression\`. The size-regression guard reverts any rewrite that fails to produce strictly smaller output.
- **Eligibility (ISO 32000-1 §7.4):**
  - No \`/Filter\`: deflate raw payload, set \`/Filter /FlateDecode\`. The headline use case — content streams imported from a parsed source PDF, which \`reader/import.go\` preserves in raw plaintext form.
  - Single \`/Filter /FlateDecode\` (without \`/DecodeParms\`): inflate, re-deflate at \`BestCompression\`. Useful when the original producer used a lower zlib level.
  - Skipped: \`/DCTDecode\` (§7.4.8), \`/JPXDecode\` (§7.4.9), \`/CCITTFaxDecode\` (§7.4.7), \`/JBIG2Decode\` (§7.4.8), multi-filter chains, malformed \`/Filter\`, FlateDecode with \`/DecodeParms\` (§7.4.4.4 predictor handling), and streams the writer is already going to compress at BestCompression (\`compress=true\`).
- **Refused on encrypted documents** — refusal runs before any mutation so a retry without the option is unaffected.
- **Composes** with \`UseXRefStream\`, \`UseObjectStreams\`, and \`OrphanSweep\`. Pass order: sweep → recompress → encrypt → serialize.

## Headline number

End-to-end on an imported text-heavy fixture: **40,569 → 6,376 bytes (84.3% saved)**. Verified text-equality round-trip across all extracted page text.

## Critical correctness fix from review

Both review subagents independently flagged that a Flate stream with \`/DecodeParms /Predictor\` (PNG/TIFF predictor per §7.4.4.4) carries predictor-filtered bytes after inflation, not plaintext. Re-deflating those bytes and dropping \`/DecodeParms\` would silently corrupt data in any reader that honors the predictor contract. Fixed by treating any FlateDecode stream with \`/DecodeParms\` as ineligible. Direct classifier test + pass-level test pin the contract.

## Other review findings addressed

- Triple-guard refusal (encrypt + UseObjectStreams + RecompressStreams) — added test pinning state-unmutated invariant.
- Orphan-dropped-before-recompress-visit ordering — added test inspecting \`w.objects\` between passes.
- Stale \`/Length\` overwrite — added test pinning that \`PdfStream.WriteTo\` overwrites a manually-set wrong \`/Length\`.
- Classifier coverage extended to non-name non-array \`/Filter\` shapes (Null, Dictionary, Boolean, Integer, String, IndirectReference), \`/Filter [/FlateDecode /FlateDecode]\` chains, and \`/Filter []\` empty arrays.
- E2E imported-document test now compares extracted text page-by-page in addition to page count + structural reparse.

## API additions

- \`document.WriteOptions.RecompressStreams bool\`
- \`core.DeflateStreamData([]byte) ([]byte, error)\` — exported (renamed from previously-unexported \`deflate\`).
- \`core.InflateStreamData([]byte) ([]byte, error)\` — symmetric inverse, the writer-side counterpart to the reader's size-bounded inflate.
- \`core.PdfStream.WillCompress() bool\` — getter matching the existing \`SetCompress(bool)\` setter, used by the recompress pass to skip streams the writer will already deflate.

## Test plan

- [x] \`go test ./...\` clean
- [x] \`go vet ./...\` clean
- [x] 28 unit tests in \`document/writer_recompress_test.go\` cover: shrink-on-raw, re-Flate-on-BestSpeed, guard reverts random bytes, all four specialized-filter skips, multi-filter skip, **predictor-bearing skip (the P0 fix)**, malformed \`/Filter\` shapes, empty/Flate-Flate arrays, encryption refusal with state invariants, triple-guard refusal, idempotence on serialized bytes, byte-identical zero-options output, composition with sweep + xref + objstm, structural reparse, round-trip-to-plaintext correctness on both raw and Flate paths, stale \`/Length\` overwrite.
- [x] 2 e2e tests in \`integration/\` round-trip an imported document and verify shrink + page-count + extracted-text equality.
- [x] \`go run ./examples/optimize\` shows the new \`+recompress\` column with 84.3% savings on the imported fixture.

### DIFF
--- a/document/writer_recompress_test.go
+++ b/document/writer_recompress_test.go
@@ -1,0 +1,820 @@
+// Copyright 2026 Carlos Munoz and the Folio Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package document
+
+import (
+	"bytes"
+	"compress/zlib"
+	"crypto/rand"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/carlos7ags/folio/core"
+)
+
+// makeFlateAt produces a Flate-encoded payload for plaintext using the
+// requested zlib level. Used by tests that need a "lower-than-Best"
+// baseline to demonstrate the recompression win.
+func makeFlateAt(t *testing.T, plaintext []byte, level int) []byte {
+	t.Helper()
+	var buf bytes.Buffer
+	zw, err := zlib.NewWriterLevel(&buf, level)
+	if err != nil {
+		t.Fatalf("zlib.NewWriterLevel(%d): %v", level, err)
+	}
+	if _, err := zw.Write(plaintext); err != nil {
+		t.Fatalf("zlib write: %v", err)
+	}
+	if err := zw.Close(); err != nil {
+		t.Fatalf("zlib close: %v", err)
+	}
+	return buf.Bytes()
+}
+
+// addRawStream wires a raw (uncompressed, no /Filter) stream into a
+// minimal Writer. Mirrors the shape produced by reader/import.go for
+// imported pages whose source was compressed by the resolver.
+func addRawStream(t *testing.T, w *Writer, data []byte) (*core.PdfStream, *core.PdfIndirectReference) {
+	t.Helper()
+	s := core.NewPdfStream(data)
+	ref := w.AddObject(s)
+	// Wire it under /Root so OrphanSweep would not drop it.
+	root, ok := w.objects[0].Object.(*core.PdfDictionary)
+	if !ok {
+		t.Fatalf("setup: first object is not the root dict")
+	}
+	root.Set("Sample", ref)
+	return s, ref
+}
+
+// addFlateStream wires a pre-Flate-compressed stream (with /Filter
+// /FlateDecode set, compress=false) into the Writer. The level
+// parameter controls how aggressively the source compressed it.
+func addFlateStream(t *testing.T, w *Writer, plaintext []byte, level int) (*core.PdfStream, *core.PdfIndirectReference) {
+	t.Helper()
+	encoded := makeFlateAt(t, plaintext, level)
+	s := core.NewPdfStream(encoded)
+	s.Dict.Set("Filter", core.NewPdfName("FlateDecode"))
+	ref := w.AddObject(s)
+	root, ok := w.objects[0].Object.(*core.PdfDictionary)
+	if !ok {
+		t.Fatalf("setup: first object is not the root dict")
+	}
+	root.Set("Sample", ref)
+	return s, ref
+}
+
+func TestRecompressStreams_NoOpOnEmptyWriter(t *testing.T) {
+	w := NewWriter("1.7")
+	w.recompressStreams() // must not panic on zero objects
+	if len(w.objects) != 0 {
+		t.Errorf("empty writer gained %d objects", len(w.objects))
+	}
+}
+
+func TestRecompressStreams_SkipsWillCompressStreams(t *testing.T) {
+	// A stream marked compress=true is what NewPdfStreamCompressed
+	// returns; the writer's WriteTo will deflate it at BestCompression
+	// already. Recompression should leave it alone — re-running here
+	// would just duplicate work for identical output.
+	w := minimalCatalogWriter(t)
+	plaintext := bytes.Repeat([]byte("compressible "), 200)
+	s := core.NewPdfStreamCompressed(plaintext)
+	originalData := s.Data
+	originalCompress := s.WillCompress()
+	ref := w.AddObject(s)
+	root := w.objects[0].Object.(*core.PdfDictionary)
+	root.Set("Sample", ref)
+
+	w.recompressStreams()
+
+	if !bytes.Equal(s.Data, originalData) {
+		t.Error("compress=true stream was rewritten")
+	}
+	if s.WillCompress() != originalCompress {
+		t.Error("compress flag flipped on a compress=true stream")
+	}
+	if s.Dict.Get("Filter") != nil {
+		t.Error("compress=true stream got a /Filter set prematurely")
+	}
+}
+
+func TestRecompressStreams_ShrinksRawStream(t *testing.T) {
+	// A raw stream (no /Filter, compress=false) is the shape produced
+	// by reader/import.go when it copies a stream that the resolver
+	// already inflated. Recompression must Flate the payload, set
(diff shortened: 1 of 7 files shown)


document/writer_recompress_test.go
```diff
@@ -0,0 +1,820 @@
...
+	// is producer-dependent but real for sufficiently large payloads.
+	w := minimalCatalogWriter(t)
+	plaintext := bytes.Repeat([]byte("re-Flate me at higher effort "), 500)
+	s, _ := addFlateStream(t, w, plaintext, zlib.BestSpeed)
+	originalLen := len(s.Data)
+
+	w.recompressStreams()
+
+	if len(s.Data) >= originalLen {
+		t.Errorf("re-Flate did not shrink: was %d (BestSpeed), now %d (BestCompression)",
```
dsymonds, line 144: FYI @carlos7ags, this test consistently fails for me on my Mac, using Go 1.27 (both 1.27.0 and 1.27.1).

dsymonds: It does _not_ fail with Go 1.26.8 (darwin/arm64), so there's a change in upstream behaviour, but I don't know if the bug is with Go or with this package.

PR #212, opened 2026-04-23
html: add Options.BaseFS and Options.Client (#85)

### LINKED ISSUE #85: Changes to html.Options
Hi,

I'm the main developer of Readeck (https://readeck.org/en/) and I can't tell you how long I've been waiting for a library like Folio. Congrats for such an impressive job with so little dependencies :)

I've been trying to convert some articles from HTML to PDF and it works great but I encountered a few limitations that could be easily lifted with some API modifications.

**Describe the feature**

- Add an `Options.BaseFS` with a Go `fs.FS` value. When nil, set it to `os.Root.FS()` on `options.BasePath`. 
- Add an `Options.Client` with an `http.Client`. When nil, defaults to `http.DefaultClient`.

These two options could then be used everywhere a remote resource is fetched (either from a path or from a URL).

I'd go as far as to remove `Options.BasePath` but that's an API break and you might not want that.

**Use case**

Being able to provide a custom `fs.FS` and `http.Client` avoids a lot of base64 encoding and decoding during the process. It would also allow for CSS fonts to load (currently you can't set them as base64 and they must be on a real filesystem so you can't load them from Go embedded resources).

If that's a change you agree with, I'd be happy to give it a try and submit a PR.

Cheers!

carlos7ags:
Closes #85.

## Summary

- Adds `html.Options.BaseFS` (`fs.FS`) so relative image / font / linked-stylesheet paths can be served from any filesystem — `embed.FS`, `os.DirFS`, `fstest.MapFS`, `(*os.Root).FS()`.
- Adds `html.Options.Client` (`*http.Client`) so every HTTP fetch (images, stylesheets) goes through a caller-supplied client. Defaults to `http.DefaultClient`.
- Keeps `BasePath` working unchanged — BaseFS is opt-in, backward compat is preserved for every existing caller.

## Resolution rules

`readAsset` in `html/converter.go`:

1. Absolute paths → `os.ReadFile` (existing absolute references keep working).
2. `BaseFS != nil` → `fs.ReadFile(BaseFS, p)`; path normalized and validated with `fs.ValidPath`, so `..` traversal attempts are rejected up front.
3. `BasePath != ""` → `os.ReadFile(filepath.Join(BasePath, p))`.
4. Otherwise → `os.ReadFile(p)` (cwd-relative).

`URLPolicy` still runs before the `Client`, so a denying policy short-circuits the fetch regardless of the client supplied.

## What changed

- `html/converter.go`: Options fields, `readAsset`, `httpClientOrDefault`, `loadFontFaces` via `font.ParseFont`.
- `html/css.go`: `parseStyleBlocks` now takes `fs.FS`.
- `html/converter_image.go`: `convertImage` / `convertImgSVG` use `readAsset`; `loadImage` replaced by `decodeImageBytes` + `converter.loadLocalImage`.
- `html/converter_helpers.go`: background-image path uses `loadLocalImage`.
- `html/fetch_image.go`: `makeCSSFetcher` and `fetchImage` use the configured `*http.Client`.
- `html/fetch_image_wasm.go`: stub signature updated.

## Out of scope

- C ABI unchanged. `fs.FS` and `*http.Client` are Go-only interfaces and can't cross the C boundary. C callers continue to use `BasePath`. Header stays in sync at 388 exports.
- `tmpl.RenderFile` still sets only `BasePath`. Candidate follow-up.
- `@font-face` inside a linked stylesheet still resolves relative to the document, not the stylesheet. Pre-existing behavior; would be a semantic change.

## Test plan

- [x] `go test ./...` (all packages)
- [x] `go test -race ./html/ -run 'TestBaseFS|TestClient'`
- [x] `GOOS=js GOARCH=wasm go build ./html/ ./layout/ ./font/ ./image/`
- [x] New test file `html/options_fs_client_test.go` covers: image via BaseFS, linked stylesheet via BaseFS, background-image via BaseFS, @font-face via BaseFS (asserted via counting FS), BaseFS precedence over BasePath, absolute path falls through to OS, custom Client for images and stylesheets, URLPolicy short-circuits a custom Client, `..` traversal rejected and falls back to alt text, direct unit test of `readAsset` path normalization and `fs.ErrNotExist` propagation, legacy BasePath still works without BaseFS.

### DIFF
--- a/html/converter.go
+++ b/html/converter.go
@@ -5,7 +5,11 @@ package html
 
 import (
 	"fmt"
+	"io/fs"
 	"math"
+	"net/http"
+	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -21,7 +25,19 @@ type Options struct {
 	// DefaultFontSize is the root font size in points (default 12).
 	DefaultFontSize float64
 	// BasePath is the base directory for resolving relative image/font/CSS paths.
+	// Used when BaseFS is nil. Paths are forwarded with filepath.Join, so
+	// OS-native separators and absolute references both work.
 	BasePath string
+	// BaseFS is the filesystem used to resolve relative paths for images,
+	// fonts, and linked stylesheets. When non-nil, it takes precedence over
+	// BasePath for relative paths. Absolute paths still fall through to the
+	// OS filesystem so existing absolute path references keep working.
+	// System-font fallback lookup in FallbackFontPath / hardcoded candidates
+	// is unaffected and always reads from the OS filesystem.
+	// Typical uses: embed.FS for embedded assets, fstest.MapFS in tests,
+	// os.DirFS(dir) or (*os.Root).FS() for sandboxed directory access.
+	// BaseFS paths must be forward-slash separated (fs.FS convention).
+	BaseFS fs.FS
 	// PageWidth is the page width in points (default 612 = US Letter).
 	PageWidth float64
 	// PageHeight is the page height in points (default 792 = US Letter).
@@ -35,6 +51,9 @@ type Options struct {
 	// Return nil to allow the fetch, or an error to block it.
 	// If nil, all URLs are allowed.
 	URLPolicy URLPolicy
+	// Client is the HTTP client used to fetch remote images and stylesheets.
+	// If nil, http.DefaultClient is used.
+	Client *http.Client
 }
 
 // URLPolicy controls whether the HTML converter may fetch a remote URL.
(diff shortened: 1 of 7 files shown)


carlos7ags:
Behavior change worth flagging: one incidental bug fix sneaked in with this PR.

`Options.defaults()` previously did not copy `FallbackFontPath` from input to output, so any caller setting `Options.FallbackFontPath` had it silently dropped — `getFallbackFont` always saw an empty string and fell through to the hardcoded system-font candidates. The new `defaults()` copies the field, so `FallbackFontPath` is now actually honored.

Impact:
- Callers who set a valid path expecting it to be used — get the behavior they already expected.
- Callers who set a broken path expecting the silent fallback to take over — will now see font loading fail. This case is unlikely (you would have to have set an invalid path and relied on the bug).

Happy to split this out into a separate commit/PR if we want this one kept strictly scoped to #85.

PR #217, opened 2026-04-24
layout: apply Unicode NFC normalization at the layout entry point

carlos7ags:
## Summary

User-supplied strings that arrive in a canonically decomposed form (for
example `U+0065 U+0301` for `e` followed by a combining acute accent)
are canonically equivalent to their precomposed form (`U+00E9`). The
shaping pipeline and the text measurers in this package expect
composed input: some font cmap tables only cover precomposed
codepoints, and width accumulation over combining marks double-counts
glyph advances that should be zero-width.

This change applies Unicode Normalization Form C (NFC) once, at the
layer boundary where user strings first become layout input.
Everything below that boundary - shaping, bidi, measurement, the draw
path - stays agnostic to the composition form of its input, and the
output for canonically equivalent strings is byte-identical.

## Why NFC at the layout boundary

- `font/` is deliberately untouched: font loaders should not need to
  know about user-facing normalization concerns, and several callers
  (imported PDFs, font parsing) rely on them being input-faithful.
- Normalizing inside `layout` covers every entry point where user
  strings become measured/shaped content while leaving the rest of
  the library alone.
- `content/` and `core/` sit below the layer and do not see user
  strings; their inputs are already glyph IDs or PDF tokens.
- NFC is idempotent, so nested entry points that call one another
  (for example `List.AddItem` eventually building a Paragraph) are
  safe even when both layers normalize.

## Entry points normalized

- `NewRun`, `NewRunEmbedded`
- `NewParagraph`, `NewParagraphEmbedded`, `NewStyledParagraph`,
  `Paragraph.AddRun`
- `Heading.SetRuns`
- `Row.AddCell`, `Row.AddCellEmbedded`
- `TabbedLine.SetSegments`

`List` items reach the pipeline through the Paragraph constructors
and are therefore covered transitively.

## Tests added (`layout/normalize_test.go`)

- `TestNormalizeTextNFC` - direct unit tests of the helper across
  empty, ASCII, already-NFC, decomposed, mixed, and combining-diaeresis
  inputs, including idempotence on repeated application.
- `TestNormalizeTextPassthroughWhenAlreadyNFC` - guards the fast path
  that returns already-NFC input unchanged.
- `TestParagraphNFCAtConstructor`, `TestNewStyledParagraphNFC`,
  `TestAddRunNFC`, `TestNewRunNFC` - every entry point normalizes
  its input.
- `TestParagraphNFCByteIdenticalLayout` - a decomposed and a
  precomposed `"cafe"` produce the same word count, same Text, same
  Width, and same `OriginalText` out of `Paragraph.Layout`.
- `TestParagraphNFCByteIdenticalDrawOutput` - the full draw path
  (`drawTextLine`) emits byte-identical content streams for the
  two inputs.
- `TestParagraphAlreadyNFCPassthrough` - already-NFC input passes
  through the constructors with no byte-level change.
- `TestParagraphMixedPartialNormalization` - strings that are only
  partially decomposed are composed in place while the rest is
  untouched.
- `TestArabicActualTextStillEmitted` - regression guard: NFC (a no-op
  for already-NFC Arabic) does not interfere with the existing
  `/ActualText` fallback used by the Arabic shaper.
- `TestTableCellNFC`, `TestHeadingSetRunsNFC`,
  `TestTabbedLineSetSegmentsNFC` - each non-paragraph entry point
  normalizes its input.
- `TestNFCTestStringsIntegrity` - guards the decomposed/precomposed
  string literals in this file against editor-level normalization.

## Tradeoffs to scrutinize

- **Always on, not opt-in.** NFC is semantics-preserving for shaping
  and is idempotent, so a flag would add surface area without a
  user-facing benefit. Reviewers should flag any scenario where a
  caller would reasonably want to preserve decomposed input.
- **Normalization cost on the hot path.** Every constructor now calls
  `norm.NFC.IsNormalString` once. For already-NFC input (the common
  case) this is a fast scan with no allocation; for decomposed input
  it allocates a new composed string once per constructor call.
- **Dependency choice.** `golang.org/x/text` is already a direct
  dependency for the Unicode bidi algorithm, so `unicode/norm` adds
  no new module and no new transitive dependencies.
- **No exported `Normalize`.** The helper is unexported on the
  premise that NFC is an internal concern of the layout pipeline.
  Exporting it would commit the API.

## Test plan

- [x] `go test ./...` passes locally
- [x] All new tests in `layout/normalize_test.go` pass
- [x] Existing Arabic `/ActualText` behavior is preserved

### DIFF
--- a/layout/element.go
+++ b/layout/element.go
@@ -164,12 +164,12 @@ type TextRun struct {
 
 // NewRun creates a TextRun with a standard font.
 func NewRun(text string, f *font.Standard, fontSize float64) TextRun {
-	return TextRun{Text: text, Font: f, FontSize: fontSize}
+	return TextRun{Text: normalizeText(text), Font: f, FontSize: fontSize}
 }
 
 // NewRunEmbedded creates a TextRun with an embedded font.
 func NewRunEmbedded(text string, ef *font.EmbeddedFont, fontSize float64) TextRun {
-	return TextRun{Text: text, Embedded: ef, FontSize: fontSize}
+	return TextRun{Text: normalizeText(text), Embedded: ef, FontSize: fontSize}
 }
 
 // RunInline creates a TextRun that represents an inline element.
--- a/layout/heading.go
+++ b/layout/heading.go
@@ -70,15 +70,11 @@ func NewHeadingEmbedded(text string, level HeadingLevel, ef *font.EmbeddedFont) 
 	}
 }
 
-// SetRuns replaces the heading's paragraph runs with the given styled runs.
+// SetRuns replaces the heading's paragraph runs with the given styled
+// runs. The runs argument is not mutated: NFC normalization is performed
+// on a fresh copy so callers retain ownership of their input slice.
 func (h *Heading) SetRuns(runs []TextRun) *Heading {
-	for i := range runs {
-		if runs[i].InlineElement != nil || runs[i].IsLineBreak {
-			continue
-		}
-		runs[i].Text = normalizeText(runs[i].Text)
-	}
-	h.para.runs = runs
+	h.para.runs = normalizeRuns(runs)
 	return h
 }
 
@@ -98,8 +94,12 @@ func (h *Heading) SetBookmarkLevel(level int) *Heading {
 
 // SetBookmarkLabel overrides the heading text used in the bookmark tree.
 // Empty string means use the heading's text content.
+//
+// The label is NFC-normalized for consistency with other layout-boundary
+// inputs. Bookmark labels are PDF metadata, not flowed text, but
+// normalizing here keeps the API uniform.
 func (h *Heading) SetBookmarkLabel(label string) *Heading {
-	h.bookmarkLabel = label
+	h.bookmarkLabel = normalizeText(label)
 	return h
 }
 
--- a/layout/heading.go
+++ b/layout/heading.go
@@ -72,6 +72,12 @@ func NewHeadingEmbedded(text string, level HeadingLevel, ef *font.EmbeddedFont) 
 
 // SetRuns replaces the heading's paragraph runs with the given styled runs.
 func (h *Heading) SetRuns(runs []TextRun) *Heading {
+	for i := range runs {
+		if runs[i].InlineElement != nil || runs[i].IsLineBreak {
+			continue
+		}
+		runs[i].Text = normalizeText(runs[i].Text)
+	}
 	h.para.runs = runs
 	return h
 }
--- a/layout/list.go
+++ b/layout/list.go
@@ -113,19 +113,27 @@ func (l *List) SetMarkerFontSize(size float64) *List {
 
 // AddItem adds a text item to the list.
 func (l *List) AddItem(text string) *List {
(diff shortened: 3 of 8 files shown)


carlos7ags:
Punch list addressed (review feedback from PR #217). New head: a0c9ea2.

Code fixes
1. List leak: `AddItem`, `AddItemWithSubList`, `AddItemRuns`, `AddItemRunsWithSubList` now NFC-normalize stored text/runs. `MinWidth`/`MaxWidth` measure on the canonical form.
2. Caller-slice mutation: `NewStyledParagraph` and `Heading.SetRuns` no longer write through the caller's slice. Both go through a new internal `normalizeRuns` helper that allocates a fresh `[]TextRun`.
12. Dropped the redundant `s == ""` short-circuit in `normalizeText`. `IsNormalString("")` is true and the fast path is allocation-free.
13. `Heading.SetBookmarkLabel` now NFC-normalizes the label for layout-boundary uniformity.

Test fixes
3. `TestArabicDecomposedNormalizedBeforeShaping` added: alef + combining hamza-above (U+0627 U+0654) feeds the Arabic path; asserts the run text and `OriginalText` are the precomposed U+0623.
4. `TestParagraphEmbeddedNFCByteIdenticalLayout` and `TestParagraphEmbeddedNFCByteIdenticalDrawOutput` added: two canonically equivalent inputs through `NewParagraphEmbedded` on the existing mock Arabic embedded font produce identical Word streams and identical draw output. No new font binaries.
5. `TestNormalizeTextVietnamese3Stack`: `ca\u0302\u0300` -> `c\u1EA7`.
6. `TestNormalizeTextHangulJamo`: `\u1100\u1161\u11A8` -> `\uAC01`.
7. `TestNormalizeTextSupplementaryPlane`: `U+1D49C` round-trips byte-identically (length stays 4).
8. `TestNormalizeTextInvalidUTF8`: pins the contract — `\xff\xfe abc` does not panic and is preserved unchanged. Documented in the helper.
9. `TestNormalizeTextMixedScripts`: Latin (decomposed) + Arabic + Devanagari. Asserts the precomposed Latin and verifies all three Unicode-block markers survive for the script splitter.
10. `TestListAddItemNFC`, `TestListAddItemRunsNFC`, `TestListMaxWidthSeesNFC`: covers the public List API and the measurement path that was the original concern.
11. Replaced the `"\u00e9"` magic with a named constant; added a nonzero-length assertion on the decomposed branch in `TestParagraphNFCByteIdenticalDrawOutput`.

Verification
- `go test ./...` clean.
- `go test -count=3 ./layout/` clean.

Decisions
- Invalid UTF-8 behavior: pass-through (`norm.NFC`'s contract); pinned by `TestNormalizeTextInvalidUTF8` and documented in the `normalizeText` doc comment so a future swap to a stricter validator must be deliberate.
- For lists, normalization happens at the public `AddItem*` boundary (i.e. on `listItem.text` and on each stored run's `Text`) rather than inside `itemParagraph` or `itemText`, so `MinWidth`/`MaxWidth` see canonical text without depending on Layout.

PR #228, opened 2026-04-30
font, html: load TrueType Collections and absolute @font-face paths

### LINKED ISSUE #227: CJK fail render some Chinese Characters.
**Describe the bug**

Image

**To reproduce**
```
cd folio
go run examples/cjk/main.go
# Created cjk.pdf

start cjk.pdf
```

**Expected behavior**

Image

**Environment**
- Go version: 
- OS:
   - Windows11 25H2 26200.8246
- Folio version/commit:
   - v0.7.1-35-gc736e46

**Additional context**

carlos7ags:
Fixes #227.

## Summary

Two bugs combined to make CJK rendering fail when the `@font-face` URL pointed at a system `.ttc` with `Options.BaseFS` unset (the reporter's setup on Windows with `msyh.ttc` and Ubuntu with `NotoSansCJK-Regular.ttc`).

- **`font.ParseFont` accepted the `ttcf` magic but routed the bytes to `sfnt.Parse`**, which rejects collections with `invalid single font (data is a font collection)`. Dispatch now extracts face 0 of the collection into a standalone single-font TTF (table directory rewritten to point at the new offsets) and parses that. Face 0 matches browser behavior for `url()` without a `#` fragment.
- **`html.loadFontFaces` resolved every filesystem path through `BaseFS`**, so an absolute system path silently failed when `BaseFS` was nil and the renderer fell through to the system fallback. When `BaseFS` is nil, an absolute path now bypasses asset resolution and loads from the OS, mirroring `loadFallbackFont`'s existing precedent. When `BaseFS` is set, a leading `/` continues to resolve at the BaseFS root so existing root-anchored asset behavior is preserved.

## Test plan

- [x] `font/ttc_test.go` covers `extractTTCFont` round-tripping through `sfnt.Parse`, `ParseFont`/`LoadFont` accepting TTC bytes/paths, and the validation paths (truncated header, wrong magic, out-of-range index, empty collection).
- [x] `html/issue227_test.go` converts HTML with `@font-face url('<absolute system path>')` and asserts that the rendered paragraph's embedded font's PostScriptName matches `font.LoadFont`'s return for the same path — distinguishing a real `@font-face` load from the fallback that previously masked the bug.
- [x] Verified that reverting either fix individually causes the corresponding tests to fail (with the html fix reverted: `expected embedded font, got standard font`; with the TTC dispatch reverted: `unknown font magic 0x74746366: font: unknown font format`).
- [x] Manual end-to-end: HTML with `@font-face url('/System/Library/Fonts/Hiragino Sans GB.ttc')` and Chinese/Japanese/Korean paragraphs now produces a PDF whose `BaseFont` is `HiraginoSansGB-W3` (the requested TTC), not the system fallback.
- [x] `go test ./...` green.

## Known follow-up

Very large CJK collections (STHeiti, some Noto CJK builds) may still hit `golang.org/x/image/font/sfnt`'s hardcoded `maxCmapSegments = 20000` limit. That is an upstream parser limit, orthogonal to TTC dispatch — the same font would fail in non-TTC form too. Worth a separate issue if it surfaces in the wild.


carlos7ags:
Addressed review feedback in f548bf0. All deferred items now in this PR — none rolled to follow-up.

**Code review (architecture / 32-bit safety):**
- `extractTTCFont` parses uint32 fields into `uint64` for all arithmetic against the buffer length. On 32-bit hosts, a uint32 ≥ 2^31 cast to `int` previously produced a negative value, defeating `srcOff+length > len(data)` and panicking on the subsequent slice — violating ARCHITECTURE.md §error-handling. Verified the `font` package cross-compiles to `GOARCH=386 GOOS=linux`.
- Added a total-size guard so `make([]byte, total)` cannot allocate the wrong size on hostile input.
- Added ISO/IEC 14496-22 §5 / Apple TrueType Reference citation on `ttcMagic` (spec-grounding rule).

**Test review (host-independent regression coverage):**
- Synthetic TTC fixture (`buildSyntheticTTC` / `buildSyntheticTTCDistinctBodies`) wraps a system TTF in TTC headers at test time. Removes the silent-skip on Linux CI without Noto-CJK — the platform the bug was reported on.
- `TestExtractTTCFontVersion2Header` locks the current "ignore version, read offset 12" behavior for v2 collections.
- `TestExtractTTCFontMultiFaceIndex` builds a 2-face TTC with distinct bodies and exercises non-zero face-index arithmetic plus out-of-range rejection.
- `TestExtractTTCFontHandlesHostileUint32Offsets` pins panic-free behavior under MaxUint32 / high-bit-set offset and numFonts.
- `TestExtractTTCFontRejectsBadInput` extended with `font-offset truncated`, `directory truncated`, `table beyond data`, `length wraps uint32`.
- `TestFontFaceCJKEndToEnd` mirrors the user's exact flow: Chinese text, `@font-face url('<absolute path>')`, `Options: nil`. Asserts the rendered word uses the requested embedded font *and* has non-zero advance — proving cmap returned glyphs, not `.notdef`. Skips when no CJK-capable system font parses cleanly.
- `TestFontFaceWindowsAbsolutePath` locks the Windows-path branch for when CI runs on Windows; skips on other hosts.

Full `go test ./...` green; cross-compile to 386 clean for the `font` package.

carlos7ags:
Splitting and re-scoping. The TTC-dispatch half of this PR — `font/ttc.go` and the `font.ParseFont` wiring — landed in #235 as a focused, standalone fix.

The `html/loadFontFaces` absolute-path half is held back: #229 will centralize asset resolution across ``, SVG, `<link>`, `@font-face`, and `FallbackFontPath`. The per-resolver branch in this PR would be redundant under that refactor and risks locking in an asymmetric rule that #229 then has to reconcile.

User impact of this split: with #235 merged, the `getFallbackFont` chain (which already uses `font.LoadFont` on absolute system paths) starts working correctly for `.ttc` files, so #227's PDF renders correctly. With #232 (`StrictAssets`, also merged) developers see the silent `@font-face` drop as an error during development — closing the observability gap.

Closing this PR. Will re-open the absolute-path piece as part of #229 if it isn't naturally absorbed.

PR #234, opened 2026-04-30
html: add Options.StrictAssets to promote asset-load failures to errors

### LINKED ISSUE #232: html: surface @font-face / image / stylesheet load failures by default
## Problem

`Options.Logger` is opt-in (`nil` by default) and `loadFontFaces`, `fetchImage`, and `parseStyleBlocks` swallow all asset-load failures into warn-level events. A user passing a wrong path or hitting the bug from #227 sees a wrong-looking PDF and has no diagnostic surface to point at the cause. The reporter of #227 had to file an issue rather than self-diagnosing.

This is a usability problem: the library's silent-recovery behavior is well-suited for production (don't crash a 1000-page report because one image 404'd) but actively hostile to development.

## Proposal

Pick one (or both):

**A. Default `Logger` to a useful sink during development.** When `Options.Logger` is nil, default to `slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: slog.LevelWarn}))` so warnings reach the user. Production users can pass `slog.New(slog.NewTextHandler(io.Discard, nil))` to silence. Requires a documented break (someone is relying on the silent default).

**B. Add `Options.StrictAssets bool`.** When true, asset-load failures return an error from `Convert` / `ConvertFull` instead of being warned-and-skipped. Trivial to add, opt-in (no break), and makes \"why is my font not embedded\" answerable in the user's own development loop.

Recommend (B) as the v0 step (no break) and (A) as a follow-up for a 0.x release.

## Acceptance criteria

- [ ] A user with a typo'd `@font-face url()` finds out about it without filing a GitHub issue
- [ ] Production behavior is preserved or opt-out
- [ ] Documented in the `Options` field doc comment and CHANGELOG

carlos7ags:
Closes #232.

## Summary

Adds a single boolean `Options.StrictAssets` that escalates the existing warn-and-continue paths to returned errors. When true, `Convert` and `ConvertFull` collect every failed `@font-face url()`, ``, `background-image: url()`, linked stylesheet, SVG load, and `FallbackFontPath` into `c.strictErrs` and return them joined via `errors.Join` at the end of the conversion. The partial result is returned alongside the error so callers can inspect both.

Defaults to false — **no breaking change**. Production keeps the warn-and-continue behavior. Use it in development and CI to surface broken asset paths in the local feedback loop instead of letting them silently degrade the output. The reporter of #227 had no diagnostic surface for their broken `@font-face`, which is the failure mode this addresses.

## Implementation

- New `(c *converter).reportAssetError` helper: always logs through `Options.Logger` (existing behavior), additionally appends a wrapped error to `c.strictErrs` when `StrictAssets` is true.
- `formatAssetError` builds the wrapped error with the slog-style attrs inlined into the message for grep-ability while preserving `errors.Is` on the underlying err via `fmt.Errorf %w`.
- Stylesheet path runs before the converter is constructed; its load callback collects into a local slice that is attached to the converter at construction time.
- All 8 existing warn sites in `html/` replaced (`converter.go` ×2 stylesheet + `@font-face` + `FallbackFontPath`; `converter_image.go` ×2 image + SVG; `converter_helpers.go` ×2 background-image fetch + load).

## Test plan

- [x] 9 new tests in `html/issue232_test.go` covering: `@font-face` escalation, default-false preserves old behavior, stylesheet escalation, `` escalation, multiple-failure joining, URLPolicy denial not escalated (current-behavior pin), partial-result-alongside-error, `errors.Is` unwraps to underlying sentinel, `ConvertFull` surfaces errors.
- [x] Verified that reverting the strict-error return-path causes 6 of 9 tests to fail with the expected `expected error, got nil` message.

## What this is not

This is the **(B)** path from #232 — opt-in `StrictAssets`, no API break. The **(A)** path (defaulting `Logger` to `slog.Default()`) is left as a follow-up since it would be a behavior change for existing nil-Logger callers and benefits from a 0.x release boundary.


carlos7ags:
Addressed review feedback in be339bf. Both reviewers' items now in this PR — none deferred.

**Code review:**
- **Blocker fixed**: `formatAssetError` no longer interpolates user-controlled attrs into a `fmt.Errorf` format string. A path like `C:\Users\foo%bar.ttf` wrote a literal `%b` into the format string, mangled the message, and broke `errors.Is` cause-chain preservation. The message is now built as a plain string via `fmt.Fprintf` into a `strings.Builder`; only the final wrap is `fmt.Errorf("%s: %w", msg, err)`. Pinned by `TestFormatAssetErrorPercentInValueIsSafe`.
- **Log attr ordering preserved**: `reportAssetError` now appends `"error", err` last, matching the historical shape so warn-log scrapers see no diff.
- **URLPolicy contract aligned with doc**: introduced public `html.ErrURLPolicyDenied` sentinel. `fetchFontBytes`, `fetchImage`, and `makeCSSFetcher` wrap policy errors with it; `reportAssetError` excludes `errors.Is(err, ErrURLPolicyDenied)` from `c.strictErrs` while still logging through `Options.Logger`. Hard-pinned by `TestStrictAssetsURLPolicyDenialNotEscalated` (replaces the prior `t.Skip`).
- **Ordering contract documented** in the `StrictAssets` field doc: stylesheets → `@font-face` → tree-walk order, byte-stable across runs.

**Test review additions:**
- `TestStrictAssetsDefaultFalseStylesheetPath` — no-break lock for the pre-converter stylesheet path
- `TestStrictAssetsBadSVGImageEscalates` — covers converter_image.go SVG branch
- `TestStrictAssetsBadBackgroundImageEscalates` — covers both converter_helpers.go branches
- `TestStrictAssetsFallbackFontPathFailureEscalates` — covers `getFallbackFont` site with non-WinAnsi text
- `TestStrictAssetsRemoteFontHTTPFailureEscalates` — pins HTTP 500 escalates and is NOT tagged as `ErrURLPolicyDenied`
- `TestStrictAssetsCollectsMultipleFailures` — tightened: type-asserts through `Unwrap() []error`, checks `len(children) == 3` (substring match alone could be spoofed by a single concatenated error)
- `TestStrictAssetsErrorOrderingDeterministic` — runs Convert twice, asserts byte-identical `err.Error()`
- `TestStrictAssetsStillLogsWhenStrict` — captured `slog.TextHandler` asserts BOTH warn-record AND error
- `TestStrictAssetsErrorsUnwrapToFsErrNotExist` — replaces the prior tautological test with `fstest.MapFS{}` + `fs.ErrNotExist` to actually exercise the load path
- `TestFormatAssetError{PercentInValueIsSafe,OddAttrs,EmptyAttrs}` — direct unit tests
- `TestStrictAssetsReturnsPartialResultAlongsideError` — tightened to assert the rendered paragraph text, not just non-empty elems

Verified each load-bearing fix is regression-catching:
- Removing `&& !errors.Is(err, ErrURLPolicyDenied)` → `TestStrictAssetsURLPolicyDenialNotEscalated` fails
- Restoring the buggy format-string concat → `TestFormatAssetErrorPercentInValueIsSafe` fails (mangled message + lost cause)

Full `go test ./...` green. PR is now 19 strict-assets tests across 8 call sites.

Issue #229, opened 2026-04-30
html: centralize asset resolution behind a single resolveLocalAsset entry point


carlos7ags:
## Problem

Five resolvers in `html/` each invented their own rule for how `Options.BaseFS`, absolute paths, and root-anchored paths interact, and none of the rules match. Issue #227 surfaced one of these inconsistencies (`loadFontFaces` did not behave like `loadFallbackFont` for absolute paths); fixing it required adding a new asymmetry rather than fixing the underlying contract.

| Call site | Absolute path | Root-anchored `/x` | `BaseFS=nil` |
|---|---|---|---|
| `loadFallbackFont` (`converter.go:584`) | OS, bypass BaseFS | OS or BaseFS root | OS |
| `loadFontFaces` (`converter.go:385`, post-#227) | OS only when BaseFS nil | BaseFS root | OS for absolute |
| `readAsset` for images (`converter_image.go:300`) | strips `/`, BaseFS only | BaseFS root | error |
| `readAsset` for SVG (`converter_image.go:258`) | strips `/`, BaseFS only | BaseFS root | error |
| `parseStyleBlocks` for `<link>` (`converter.go:262`) | BaseFS only | BaseFS root | error |

`Options.BaseFS`'s doc claims it resolves "all local paths," but the `loadFallbackFont` carve-out already violates that. Each new asset type is a coin-flip on whether its resolution rule matches user expectations.

## Proposal

Extract a single method `(c *converter) resolveLocalAsset(origin, src string) ([]byte, error)` that every consumer (image, SVG, `<link>`, `@font-face`, `FallbackFontPath`) calls. The contract:

1. `data:` URI → decode inline
2. Absolute http/https URL → fetch via `Options.Client` subject to `URLPolicy`
3. HTTP-origin stylesheet relative → resolve against origin URL
4. Absolute filesystem path AND `BaseFS == nil` → load from OS
5. Anything else → resolve via `joinFSPath` and read through `BaseFS`, error if `BaseFS == nil`

Document the contract in a new `### Asset resolution` section in `ARCHITECTURE.md` (or `html/RESOLUTION.md`).

## Why this matters

This eliminates an entire class of "resolver X behaves differently from resolver Y" bugs, including ones not yet reported. Any future asset type (cursor URLs, list-style images, `<object>` data) inherits the contract for free.

## Out of scope

[...]

PR #235, opened 2026-04-30
font: load TrueType Collections via face-0 extraction

### LINKED ISSUE #227: CJK fail render some Chinese Characters.

carlos7ags:
Partial fix for #227. Splits the TTC-dispatch half of #228 into a focused, scope-minimal PR; the companion `loadFontFaces` absolute-path patch is held back as redundant under #229 (centralize asset resolution).

## Summary

`font.ParseFont` accepted the `ttcf` magic but routed the bytes to `sfnt.Parse`, which rejects collections with `invalid single font (data is a font collection)`. Dispatch now extracts face 0 of the collection into a standalone single-font TTF (table directory rewritten to point at the new offsets) and parses that. Selection of face 0 matches browser behavior for `url()` references without a `#` fragment.

## Why this fixes #227 on its own

The reporter's failure mode on Windows/Ubuntu came from `msyh.ttc` / `NotoSansCJK-Regular.ttc` failing to load. Two paths reach those files:

1. **`@font-face url()`** — when `Options.BaseFS` is nil, the path silently fails (issue #229 will fix this uniformly across all resolvers).
2. **`getFallbackFont` chain** — already uses `font.LoadFont` on absolute system paths directly, no `BaseFS` involvement. Once TTC dispatch works, `msyh.ttc` succeeds via the fallback path and the document renders correctly.

So this PR alone makes the user's CJK PDF render correctly. With `StrictAssets` (#232, just merged), developers running with strict mode now see the silent `@font-face` drop as an error during development — closing the observability gap the reporter hit.

## Implementation (`font/ttc.go`)

- `extractTTCFont(data, index)` parses the TTC header, validates with `uint64` arithmetic so hostile `uint32` fields cannot wrap to negative ints on 32-bit hosts (ARCHITECTURE.md §error-handling: never panic on malformed input), copies each table from its absolute TTC offset into a new buffer, and rewrites the directory entries to point at the new positions. The output is a fully self-contained single-font TTF that downstream code (`sfnt.Parse`, `parseTTFTables`, `Subset`, `embed`) handles unmodified.
- TTC v2 (12-byte DSIG fields after the offset-table array) handled implicitly because the version field is not consulted; the DSIG fields sit after the data the extractor reads.
- `ttcMagic` constant + ISO/IEC 14496-22 §5 / Apple TrueType Reference citation at its declaration.

## Test plan

- [x] `buildSyntheticTTC` / `buildSyntheticTTCDistinctBodies` fixtures wrap any system TTF in TTC headers at test time. Eliminates the silent-skip on hosts without a system `.ttc`; the only fallback is when no TTF is found at all (rare on every dev/CI host).
- [x] `TestExtractTTCFontFromSyntheticTTC` — round-trips through `sfnt.Parse`.
- [x] `TestExtractTTCFontVersion2Header` — locks "ignore version, read offset 12" for v2 collections.
- [x] `TestExtractTTCFontMultiFaceIndex` — exercises non-zero face index via a 2-face TTC with distinct bodies; verifies both faces parse and the index-out-of-range case rejects with `ErrCorruptTable`.
- [x] `TestExtractTTCFontHandlesHostileUint32Offsets` — pins panic-free behavior under MaxUint32 / high-bit-set offset and numFonts; verified the `font` package cross-compiles to `GOARCH=386 GOOS=linux`.
- [x] `TestExtractTTCFontFromRealSystemTTC` — opportunistic real-TTC validation against parseable system fonts.
- [x] `TestParseFontAcceptsTTC` / `TestLoadFontAcceptsTTC` — full `ParseFont` / `LoadFont` pipeline against synthetic TTC bytes and a temp file.
- [x] `TestExtractTTCFontRejectsBadInput` — table-driven over 9 malformed-input shapes (too-short, wrong-magic, out-of-range index, negative index, empty collection, font-offset-truncated, directory-truncated, table-beyond-data, length-wraps-uint32). Each asserts `errors.Is` matches the documented sentinel.

## Scope vs. #228

This PR is the standalone correctness half. The `loadFontFaces` absolute-path branch from PR #228 is held back: #229 will centralize asset resolution across ``, SVG, `<link>`, `@font-face`, and `FallbackFontPath`, making the per-resolver patch redundant. Closing #228 once this lands.

## Known follow-up

Very large CJK collections (STHeiti, some Noto CJK builds) may still hit `golang.org/x/image/font/sfnt`'s hardcoded `maxCmapSegments = 20000` limit. That is an upstream parser limit, orthogonal to TTC dispatch — the same font would fail in non-TTC form too. Worth a separate issue if it surfaces in the wild.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - **SVG `preserveAspectRatio` slice viewport clip** — when an SVG is drawn with `slice` meet-or-slice, the renderer now emits a PDF clip path on the target rectangle before the viewport transform. Previously the uniform scale was applied correctly but content outside the target rectangle leaked onto the page. Callers that already clipped externally will continue to work (#196)
+- **TrueType Collection (`.ttc`) fonts now load** — `font.ParseFont` and `font.LoadFont` advertised TTC support but routed the bytes to `sfnt.Parse`, which rejects collections with `invalid single font (data is a font collection)`. The dispatch now extracts face 0 from the collection into a standalone single-font TTF (table directory rewritten to point at the new offsets) and parses that. Selection of face 0 matches browser behavior for `url()` references without a `#` fragment. Very large CJK collections may still hit `golang.org/x/image/font/sfnt`'s hardcoded `maxCmapSegments` limit — that is an upstream parser limit, orthogonal to TTC dispatch (#227)
 
 ## [0.7.1] - 2026-04-22
 
--- a/font/load.go
+++ b/font/load.go
@@ -10,7 +10,9 @@ import (
 )
 
 // ParseFont parses a font from raw bytes, auto-detecting the format.
-// Supports TTF, OTF, and WOFF1 fonts.
+// Supports TTF, OTF, WOFF1, and TrueType Collection (TTC) fonts. For
+// collections, the first face is selected — matching the convention used
+// by browsers and font tools for url() references without a `#` fragment.
 //
 // Errors returned by this function wrap one of the sentinel errors
 // [ErrUnknownFormat], [ErrTruncated], or [ErrCorruptTable] so callers
@@ -27,11 +29,16 @@ func ParseFont(data []byte) (Face, error) {
 			return nil, fmt.Errorf("decode WOFF: %w", err)
 		}
 		return ParseTTF(ttfData)
+	case ttcMagic:
+		ttfData, err := extractTTCFont(data, 0)
+		if err != nil {
+			return nil, fmt.Errorf("decode TTC: %w", err)
+		}
+		return ParseTTF(ttfData)
 	case 0x00010000, // TrueType
 		0x4F54544F, // "OTTO" (OpenType/CFF)
 		0x74727565, // "true"
-		0x74797031, // "typ1"
-		0x74746366: // "ttcf" (TrueType Collection)
+		0x74797031: // "typ1"
 		return ParseTTF(data)
 	}
 	return nil, fmt.Errorf("unknown font magic 0x%08X: %w", sig, ErrUnknownFormat)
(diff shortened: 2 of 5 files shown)


carlos7ags:
Addressed the two must-have items from the test review.

**Must-have #1 — `html.TestFallbackChainResolvesTTCViaBaseFS`**: pins the user-flow claim that this PR alone makes #227 render correctly via the fallback chain. Routes through `loadFallbackFont` against a `fstest.MapFS` carrying a synthetic TTC (built test-locally from any system TTF). Stays on the relative-path branch so the test does NOT depend on the absolute-path / BaseFS-bypass behavior held back under #229. Verified the test fails with a clear `TTC dispatch broken?` diagnostic when `font.ParseFont`'s `ttcf` arm is reverted.

**Must-have #2 — `font.TestExtractTTCFontHonorsFaceIndex`**: distinguishes "extractor honored the index" from "extractor returned face 0 for any index". Note: the test reviewer's suggested `bytes.Equal(face0, face1)` assertion turned out to be wrong — when both faces wrap the same TTF, the extracted single-font buffers are byte-identical because directory offsets are rewritten relative to each new buffer's origin (each starts at 0). I took a different tack: build a 2-face TTC where face 0 points at valid bytes and face 1's offset is past EOF. A correct extractor succeeds on index 0 and rejects index 1 with `ErrTruncated`; a buggy extractor that ignores the index would succeed on both. Verified the test fails when `extractTTCFont` is short-circuited to ignore the index argument.

Nice-to-have items (TTC v2 with non-zero DSIG, shared-table coverage, `numTables = 0xFFFF`, embedded testdata under `font/testdata/`) were marked out-of-scope by the same test review and are left for follow-up.

Full `go test ./...` green; rebased on top of latest `main`.

PR #239, opened 2026-04-30
html: centralize asset resolution behind one resolveLocalAsset

### LINKED ISSUE #229: html: centralize asset resolution behind a single resolveLocalAsset entry point

carlos7ags:
Closes #229.

## Summary

Every local or remote asset reference in the `html` package — ``, inline SVG, `<link rel="stylesheet">`, `@font-face url()`, `background-image: url()`, and `Options.FallbackFontPath` — previously carried its own routing logic that disagreed on absolute paths, root-anchored paths, and HTTP origins. They now route through one method, `resolveLocalAsset`, with the rules pinned in its doc comment.

This closes the workaround originally proposed in PR #228 (`@font-face url('/abs/system/font.ttf')` with `BaseFS: nil` now succeeds via the centralized contract). It also fixes the pre-existing inconsistency where `` did not flow through `URLPolicy` enforcement the same way raster images did.

## The contract

Evaluated in this order against `(opts, urlPolicy, origin, src)`:

1. `http(s)://` `src` — fetched via `opts.Client` subject to `urlPolicy`. Denials wrap with `ErrURLPolicyDenied`.
2. `http(s)://` `origin`, relative `src` — `src` resolved as URL relative to `origin`, then fetched.
3. `filepath.IsAbs(src)` and `opts.BaseFS == nil` — OS read via `os.ReadFile`.
4. Otherwise — `joinFSPath(origin, src)` then BaseFS read; `errNoBaseFS` if BaseFS is nil and src is non-absolute.

`data:` URIs are NOT routed through the contract; each asset type has its own metadata-aware decoder.

## Implementation

- Free function `resolveLocalAsset(opts, urlPolicy, origin, src, maxBytes)` callable both from `*converter` and from `parseStyleBlocks` (which runs before the converter exists).
- Method wrapper `(c *converter).resolveLocalAsset` for post-construction call sites.
- `fetchHTTPBytes` consolidates the `URLPolicy`-then-`httpGetBytes` pattern; `ErrURLPolicyDenied` wrapping happens in one place.
- `parseStyleBlocks` signature changed from `(doc, baseFS, fetchURL, onLoadError)` to `(doc, opts, urlPolicy, onLoadError)` so it can call the resolver directly.

## FallbackFontPath carve-out

`Options.FallbackFontPath` is a programmatic option set by the embedding application — not asset reference supplied by HTML content — so two extra rules apply only to it:

1. Absolute paths always bypass BaseFS to the OS, regardless of whether BaseFS is set.
2. Relative paths that miss in BaseFS retry against the OS.

Document-supplied references do NOT get these extensions — the trust boundary is different. Carve-out documented in `loadFallbackFont`'s comment and CHANGELOG. Existing test `TestFallbackFontAbsoluteSkipsBaseFS` pins it.

## Removed (~120 LOC net delete)

- `converter.fetchFontBytes`
- `converter.fetchImage` (both `!js` and `js` stubs)
- `converter.loadLocalImage`
- `makeCSSFetcher` (both `!js` and `js` stubs)

## Test plan

- [x] Both `!js` and `js`/`wasm` builds compile.
- [x] `gofmt -s -l .` clean.
- [x] No new tests added — the centralization is behavior-preserving for every existing test (intentional). Future PRs that change resolution rules should add tests at the contract level.

## Known narrow change

The old `fetchImage` consulted the HTTP `Content-Type` header for image format detection before falling back to URL extension and content sniffing. The centralized path uses URL extension + content sniffing in `decodeImageBytes` (no access to response headers). No existing test exercised the Content-Type-vs-extension disagreement case, so this is captured as a behavior note in the commit message rather than a failing test.

## ARCHITECTURE.md

Adds a short "Asset resolution" section pointing at the `resolveLocalAsset` doc comment, plus an "Evaluating a proposal" reminder that new resolvers must route through this contract. Detailed rules live next to the code (Go doc comment) so they can't drift.


carlos7ags:
Addressed feedback from both review passes in 6ba6914.

**Code review (no blockers, 2 minor cleanups):**
- Dropped the redundant `urlPolicy` parameter from `parseStyleBlocks` — always `opts.URLPolicy` at every call site.
- Tightened the `resolveLocalAsset` doc: rule #3 now explicitly covers absolute src when BaseFS IS set (web-style root via `joinFSPath` stripping the leading slash); `maxBytes` documents the "0 or non-positive" default; one-line note that src is pre-trimmed.

**Test review (4 must-haves):**
- `TestResolverMatrixAbsolutePathNoBaseFS` — parameterized over ``, `<link rel=stylesheet>`, `background-image`, and `@font-face`. Pins rule #3 (absolute path + BaseFS nil → OS read) across all document-supplied resolvers — the new behavior the centralization introduced. The `@font-face` subtest uses a negative-path detection (missing file must produce OS-style "no such file" error, not `errNoBaseFS`) since seeding a real font fixture is out of scope.
- `TestDocumentAbsolutePathUsesBaseFSWhenSet` — pins the trust-boundary distinction: `` with BaseFS set reads from BaseFS root, NOT OS (the `FallbackFontPath` carve-out doesn't extend to document-supplied references).
- `TestSVGImgSrcGoesThroughURLPolicy` — pins the side-effect behavior the PR introduces: SVG via http(s):// now flows through `URLPolicy` uniformly with raster images.
- `TestImageFormatDetectionPrefersExtensionOverContentType` — regression pin for the acknowledged narrow change: server returns `Content-Type: image/png` for a `.jpg` URL containing JPEG bytes, must decode successfully via URL extension + magic-byte sniffing.
- `TestStrictAssetsRemoteFontHTTPFailureEscalatesNotPolicyTagged` — follow-up safety check that `ErrURLPolicyDenied` wrapping is scoped to actual policy denials, not generic HTTP failures.

Verified each load-bearing fix catches its corresponding regression by temporarily reverting rule #3 (`case opts.BaseFS == nil && filepath.IsAbs(src)`); all four matrix subtests fail with clear `errNoBaseFS` diagnostics, including the negative-path `@font-face` case.

Full `go test ./...` green; both `!js` and `js` (WASM) builds clean; `gofmt` clean.

Nice-to-haves deferred per the test reviewer's own scope guidance: stylesheet 10MB cap regression test, URLPolicy-denial parameterized across all resolvers, contract-matrix synthesis test, WASM CI step.

Issue #231, opened 2026-04-30
examples: ship without CI; docs and code drift apart silently


carlos7ags:
## Problem

`examples/` has no `*_test.go` files. The example programs ship with detailed doc comments listing required system fonts and platform-specific expectations, but nothing exercises them end-to-end. Issue #227 was effectively the user being our CI for `examples/cjk/main.go`:

- The example's doc comment said it works on Windows with `msyh.ttc` and Ubuntu with `NotoSansCJK-Regular.ttc`
- Both paths hit the TTC-dispatch bug
- The example was tested on macOS, where `findCJKFont()` happens to return `Arial Unicode.ttf` (a TTF, not a TTC)
- Nobody caught it until a Windows user ran it

## Proposal

Add a test per example that:

1. Compiles and runs the example with a synthetic / minimal-system fixture (reuse the `loadAnySystemTTF` + `buildSyntheticTTC` helpers that landed for #227)
2. Asserts the produced PDF starts with `%PDF-` and embeds the requested font (BaseFont check)
3. Skips cleanly when no compatible fixture is available, rather than skipping silently the whole thing

Sketch:

```go
// examples/cjk/main_test.go
func TestCJKExampleProducesValidPDF(t *testing.T) {
    if testing.Short() { t.Skip(\"example test (use -tags examples to enable in default runs)\") }
    // Stub findCJKFont to return a synthetic TTC built from a Latin TTF, so
    // the test runs on any host. Or use os.Setenv to point findCJKFont at
    // a fixture path.
    // Run main(), capture stdout, assert cjk.pdf starts with %PDF- and
    // includes the requested @font-face PostScript name as BaseFont.
}
```

Some examples may need a small refactor (extract `buildHTML` + `findCJKFont` from `main()` so tests can call them directly without re-running `os.Exit`).

## Scope suggestion

Phase 1: `examples/cjk` (the one #227 reported). Phase 2: roll out to other examples on the same template (`examples/fonts`, `examples/zugferd`, `examples/sign`).

## Acceptance criteria

- [ ] At least one example has an integration test that would have caught the broken path in #227
- [ ] CI runs the example tests on every push
- [ ] A pattern others can copy for new examples

carlos7ags:
[...]

PR #245, opened 2026-04-30
examples/cjk: add end-to-end test (Phase 1 of #231)

carlos7ags:
Phase 1 of #231.

## Summary

`examples/cjk` previously had no test coverage. The doc comments listed working Windows / Linux font paths that were never exercised end-to-end because nothing in CI compiled or ran the example. Issue #227 was the consequence — the reporter became the de-facto CI.

This PR adds two tests next to the example so it ships with its own regression-pin.

### `TestCJKExampleProducesValidPDF`
Builds the example's HTML against a synthetic TTC fixture (constructed at test time by wrapping any system TTF in a TTC envelope), runs `html.ConvertFull` and `document.Save`, and asserts:
- The resulting bytes start with `%PDF-`.
- The PDF size is at least 1KB (rejects empty / error documents).
- **The PDF embeds the requested font specifically** — the assertion looks for `+<PostScriptName>` (the 6-character subset prefix Folio writes before the BaseFont name). A weaker `/BaseFont`-presence assertion was rejected during development: when `buildHTML` was simulated to drop `@font-face`, the PDF still contained `/BaseFont` because the system fallback embedded itself. The `+<PostScriptName>` needle catches that — verified the test fails with `looking for "+ArialMT"` when `@font-face` is removed.

### `TestCJKExampleFindCJKFontReturnsExistingPath`
Verifies that when `findCJKFont` returns non-empty, the path exists on disk. macOS major versions, Linux distribution changes, and Windows font set drift will surface here as a soft failure rather than at user runtime.

## Test plan

- [x] Both tests pass on macOS.
- [x] `TestCJKExampleProducesValidPDF` verified to fail with a clear diagnostic (`looking for "+ArialMT"`) when `buildHTML` is simulated to drop the `@font-face` rule.
- [x] Full `go test ./...` green (the example test runs as part of `./examples/cjk/`).
- [x] `gofmt -s -l .` clean.

## Not in this PR (Phase 2)

The same pattern needs to roll out to other examples that depend on system assets: `examples/fonts`, `examples/zugferd`, `examples/sign`, etc. Each will need ~30 LOC of `_test.go` next to its `main.go`. Tracking under #231 once the pattern lands.

## Implementation notes

- The synthetic TTC fixture is duplicated from `font/ttc_test.go`'s `buildSyntheticTTC` pattern (~25 LOC). The original is unexported and the example test cannot call it. Cheap duplication, no public API surface change. Could be promoted to a shared test helper later if more examples need the same shape.
- The TTF candidate list mirrors the universal set used by the font package's test fixtures so the example test runs on the same hosts the unit tests do.
- The test skips when no system TTF is available (rare on dev/CI hosts; covers minimal Alpine images without any font package installed).


carlos7ags:
Addressed feedback from both review passes in db02042.

**Test review (must-have):**
- `TestCJKExampleQuotesFontURL` — unit-level guard for the CSS src format. Without this, a regression that drops the quotes (`url(%s)` instead of `url('%s')`) could silently produce a valid-looking PDF via the system fallback chain since the `+<PostScriptName>` needle wouldn't reliably catch it on tempdir paths with no special characters. Verified the test fails with a clear `"did not emit single-quoted url('...')"` diagnostic when quotes are removed from `buildHTML`.

**Code review (optional nit):**
- One-paragraph note added to the `+<PostScriptName>` assertion documenting that `font/embed.go` runs the name through `sanitizePSName` before the subset prefix is glued on. The sanitizer is a no-op for every TTF in `loadAnyTTF` today (`ArialMT`, `DejaVuSans`, `LiberationSans-Regular`, `NotoSans-Regular`). If a future candidate has characters the sanitizer rewrites, the assertion needs to mirror that transformation.

**Deferred per the test reviewer's own scope guidance:**
- Full PDF structural parse via the reader package — nice-to-have, out-of-scope until Phase 2.
- Per-OS `findCJKFont() != ""` assertion gated by env var — nice-to-have.
- Pre-extracting `wrapTTFAsTTC` into `examples/internal/testfixtures` — revisit when Phase 2 has a second example.

Full `go test ./...` green; `gofmt` clean.

PR #251, opened 2026-05-01
font: own the cmap parser to load fonts beyond sfnt's 20000-segment limit

### LINKED ISSUE #227: CJK fail render some Chinese Characters.

### LINKED ISSUE #248: font: large CJK fonts hit golang.org/x/image/sfnt's hardcoded maxCmapSegments=20000 limit
## Problem

After #235 fixed TTC dispatch in `font.ParseFont`, large CJK font files still fail to load with:

```
sfnt: unsupported number of cmap segments
```

The error originates from `golang.org/x/image/font/sfnt/sfnt.go:69` (currently at v0.39.0):

```go
maxCmapSegments = 20000
```

Many CJK fonts have more cmap segments than this hardcoded ceiling because they cover tens of thousands of CJK Unified Ideographs. This affects the fonts most commonly referenced in the wild for East-Asian text:

| Font | Platform | Cmap segments | Loads? |
|---|---|---|---|
| `msyh.ttc` (Microsoft YaHei) | Windows | exceeds 20000 | ❌ |
| `NotoSansCJK-Regular.ttc` | Linux (some packages) | exceeds 20000 | ❌ |
| `STHeiti Light.ttc` | macOS | exceeds 20000 | ❌ |
| `Hiragino Sans GB.ttc` | macOS | within limit | ✅ |
| `Helvetica.ttc` | macOS | within limit | ✅ |
| Most non-CJK TTC | any | within limit | ✅ |

The `font` package's TTC dispatch routes correctly; the failure is downstream in `sfnt.Parse`. So this is orthogonal to TTC handling — the same font rejected as a `.ttc` would also be rejected as a `.ttf` if extracted.

## User-visible impact

Issue #227 was reported on Windows (msyh.ttc) and Ubuntu (NotoSansCJK-Regular.ttc). PR #235 made TTC dispatch work, but for these specific fonts the user still sees the same broken-CJK output because `sfnt.Parse` rejects the bytes after the dispatch succeeds. The example program (`examples/cjk`) avoids this on macOS only because `findCJKFont()` happens to prefer `Arial Unicode.ttf` first, which is below the limit.

## Workarounds users have today

- Switch to a smaller CJK font (Hiragino Sans GB, Source Han Sans subsets, region-specific variants).
- Pre-subset the font offline before passing it to Folio.
- Convert the offending TTC to TTF and strip unused cmap subtables with a tool like `fonttools`.

None of these work for a user who just wants `` to render their resume.

### A. File upstream and wait
The sfnt limit is in `golang.org/x/image`. Has been there since at least v0.19.0. Filing a Go issue and bumping when fixed is the cleanest long-term path, but timing is unpredictable.

### B. Vendor a patched sfnt
Fork the package (or the relevant cmap.go file) with a higher limit. Adds a dependency-policy violation per ARCHITECTURE.md.

### C. Implement our own cmap parser, bypass sfnt for the cmap table
Build a CMAP-format-12 reader in `font/`, detect when the source has too many segments, parse it ourselves, and feed sfnt a synthesized lower-segment-count cmap. Substantial work but possible — the rest of the font infrastructure (subset, embed, glyph IDs) is independent.

### D. Switch to a different sfnt parser
There are other Go font parsers (e.g., `github.com/go-text/typesetting/font`). A wholesale switch is a large change with cascading test updates.

### E. Document the limitation, recommend pre-subsetting
The pragmatic short-term move. Add a section to the user docs about `maxCmapSegments` and provide a `fonttools` recipe.

## Recommendation

**E now, A or C later.** Document the limitation immediately so users hitting this know it's not their fault. File the upstream issue. Revisit when there's enough demand to justify (C).

## Relevant references

- `font/ttc.go` (the dispatch)
- `font/truetype.go` line ~58 (the `sfnt.Parse` call)
- CHANGELOG entry under #227 already mentions this limit
- `golang.org/x/image/font/sfnt/sfnt.go` line 69 (`maxCmapSegments = 20000`)
- `golang.org/x/image/font/sfnt/cmap.go` lines 143, 266 (the rejection sites)

## Acceptance criteria

- [ ] Document the limitation in user-facing docs (CHANGELOG callout, README, or `font` package doc)
- [ ] File the upstream issue against `golang.org/x/image`
- [ ] Decide on long-term path (B/C/D) once data on user impact arrives

carlos7ags:
Closes #248. Fixes #227 end-to-end for the fonts the original reporter was using.

## What this fixes

`golang.org/x/image/font/sfnt` has a hardcoded `maxCmapSegments = 20000` constant. Microsoft YaHei (`msyh.ttc`), Noto Sans CJK (`NotoSansCJK-Regular.ttc`), and STHeiti all declare 23k–30k cmap groups in their format-12 subtable — comfortably above the limit. After PR #235 fixed TTC dispatch, the user-reported #227 scenario was still broken because the same fonts hit this orthogonal sfnt limit downstream.

Pre-flight measured the actual segment counts:

| Font | sfnt.Parse | format-12 numGroups |
|---|---|---|
| STHeiti Light | FAIL: cmap segments | 23,495 |
| STHeiti Medium | FAIL: cmap segments | 23,418 |
| Hiragino Sans GB | OK | 14,031 (under limit) |
| Arial Unicode | OK | 310 (format 4) |

Confirmed: only the cmap limit blocks. No other sfnt rejections. Phase 1 (this PR) alone fixes the user's scenario.

## How it works

Two new files:

- **`font/cmap.go`** — `parseCmapTable` reads the cmap directly from raw font bytes. Supports format 4 (BMP-only) and format 12 (full-Unicode). Subtable picker prefers format-12 Unicode-platform encodings over BMP-only fallbacks.
- **`font/cmap_stub.go`** — `appendStubCmap` copies the font bytes and rewires the directory's cmap entry to point at a 22-byte minimal stub appended at the end. Original cmap stays orphaned in the buffer (harmless). sfnt parses the rest of the font using the stub; Folio supplies the real mapping via `GlyphIndex`.

`ParseTTF` gains a fallback branch that detects the sfnt error message and routes through the recovery path. `sfntFace.GlyphIndex` consults `folioCmap` when set, falling through to sfnt for the existing fast path on small fonts. No behavior change for fonts that were already loading.

## Why not just bump sfnt's limit upstream

The upstream timing is unpredictable (the limit has been there for years). The recovery path is small enough (~280 LOC of new code, all straightforward binary parsing) to absorb without forking. The upstream issue can still be filed separately; this PR doesn't depend on the outcome.

## Test plan

- [x] `TestParseCmapFormat4Synthetic` — format-4 round trip
- [x] `TestParseCmapFormat12LargeNumGroups` — synthetic 25000-group fixture (above sfnt's limit), pinned regression for the user's actual fonts
- [x] `TestParseCmapSubtableSelection` — format-12 wins over format-4
- [x] `TestParseCmapMissingTable` — error path with `errors.Is(err, ErrMissingTable)`
- [x] `TestAppendStubCmapRedirects` — directory rewires to the stub at a new offset, original cmap orphaned but preserved
- [x] `TestSTHeitiLoadsViaRecoveryPath` — real-system test against macOS STHeiti Light (proxy for the user's `msyh.ttc` / `NotoSansCJK-Regular.ttc`); verified the test fails with `sfnt: unsupported number of cmap segments` when the recovery branch is reverted; passes with the recovery in place AND returns a valid GID for U+4E2D '中' (the first character of the user's reproduction text)
- [x] End-to-end smoke before commit: the user's scenario (`@font-face url(absolute-STHeiti-path)` + Chinese text + `BaseFS: nil`) produces a PDF with `BaseFont /<subset>+STHeitiTC-Light` — the requested font correctly embedded
- [x] Full `go test ./...` green
- [x] Both `!js` and `js`/`wasm` builds clean
- [x] `gofmt` clean

## What this fixes for the user

Issue #227 reporter @liserjrqlxue can now run `examples/cjk` on Windows with `msyh.ttc` and on Ubuntu with `NotoSansCJK-Regular.ttc` and get a correctly-rendered Chinese PDF — matching the expected HTML rendering they posted in the original report.

## Phase 2 (out of scope, future work)

A future PR can drop `sfnt` entirely by extending the recovery path to also bypass sfnt for metrics/names/glyph data, reading head/hhea/OS/2/name/hmtx/loca/glyf directly. That unblocks any future sfnt limit without further special cases. This PR is scope-minimal — sfnt is still used for everything except cmap.


carlos7ags:
Addressed feedback from both review passes in a75585e.

**Code review (blocker fixed):**
- `scoreSubtable` had a dead `format == 4` arm under platform 0 / encoding 4 / format 4 returning 60. Encoding 4 declares "non-BMP allowed" which format 4 (BMP-only) cannot express, so no real font ships this combination. Replaced the unreachable case with a comment documenting why we don't accept it.

**Test review (3 must-haves):**
- `TestParseTTFRecoversFromOversizedCmap` — closes the Linux-CI gap (the macOS-only STHeiti test wasn't enough). Surgically injects a 25000-group format-12 cmap into any system TTF, calls `ParseTTF`, and asserts the recovered Face returns the GID we wrote into the synthetic cmap for a non-BMP probe rune. Verified the test fails with `parse font: sfnt: unsupported number of cmap segments` when the recovery branch is reverted.
- `TestAppendStubCmapAcceptedBySfnt` — confirms sfnt actually accepts the 22-byte stub. Without this, a future bump to `golang.org/x/image/font/sfnt` that tightens cmap validation could silently break recovery on every CJK font.
- `TestGlyphIndexUsesSfntWhenNoFolioCmap` — pins the fast path. Loads a small TTF and asserts `GlyphIndex('A')` is non-zero. Verified the test fails with `GlyphIndex('A') returned 0; the fast-path branch may have regressed` when the `folioCmap != nil` condition is inverted.

**Deferred per both reviewers' scope guidance** (non-blocking, ~80 LOC of nice-to-haves):
- `TestParseCmapFormat4IndirectPath` (idRangeOffset!=0 branch — needs glyphIdArray fixture; exercised by real fonts via STHeiti).
- `TestParseCmapFormat4NegativeDelta` (uint16 wraparound — code review traced the math and verified it's correct).
- `TestParseCmapIgnoresFormat14` (variation selectors stub).
- `TestScoreSubtable` table-driven across priority tiers.
- `TestParseCmapErrorCases` for malformed inputs.
- `BenchmarkParseTTFSmall` to pin the fast-path-no-cost claim.

Full `go test ./...` green; both `!js` and `js`/`wasm` builds clean; `gofmt` clean.

carlos7ags:
**Pre-merge risk audit confirmed: no silent correctness bug.** Risk #3/#10 (subset / ToUnicode CMap reading from sfnt's stub instead of `usedGlyphs`) was the only one that could have shipped a usable-looking but accessibility-broken PDF.

Verified by inspection (`font/subset.go:92` rebuilds cmap from `usedGlyphs`, `font/embed.go` builds ToUnicode CMap from the same source — both populated by `GlyphIndex` during shaping → consults our folioCmap correctly).

Verified end-to-end with a smoke test: rendered the user's exact phrase through the recovery path, re-parsed the resulting PDF via `reader.Parse`, called `page.ExtractText()` (the same path Acrobat / pdftotext / screen readers use). Output: `"中华人民共和国是一个历史悠久的文明古国。"` — byte-perfect round trip.

**Pinned** the verification as `integration/cjk_recovery_roundtrip_test.go` in 592f219. macOS-only via `runtime.GOOS` since STHeiti is the universally-available recovery-path-triggering fixture; the synthetic `TestParseTTFRecoversFromOversizedCmap` covers the unit-level wiring on every host. Verified the new integration test fails with `parse font: sfnt: unsupported number of cmap segments` when the recovery branch is reverted.

Other risks from the audit, status update:
- **#1 (orphaned cmap inflates font binary)**: real, ~300KB per CJK font load. Acceptable for v0; a follow-up could write a compacted output instead of append-stub.
- **#2 (`isUnsupportedCmapErr` substring matching is brittle)**: real, captured in PR description. Worth a CONTRIBUTING note (deferred).
- **#5 (stub format-6 acceptance by sfnt)**: pinned by `TestAppendStubCmapAcceptedBySfnt` in the previous review-fix commit.
- **#7 (Linux/Windows CI never sees recovery on real fonts)**: partially closed by `TestParseTTFRecoversFromOversizedCmap` (synthetic, runs everywhere) plus the new integration test (real font, macOS). Real-font Linux CI would need a `font/testdata/` fixture we don't ship.
- **#8 (format 14 silently dropped)**: documented in `parseCmapTable` doc comment; behavior intentional for v0.
- **#9 (DoS-resistance on hostile font claiming `numGroups=0xFFFFFFFF`)**: bounded by buffer-length check in `parseCmapFormat12`. Should add a malformed-input test in a follow-up.

All highs and mediums either resolved, pinned, or documented. Cleared to merge.

carlos7ags:
**Closing without merge — superseding with Phase 2.**

User feedback during pre-merge audit: the recovery-path approach trades structural risk (substring-matching sfnt's error message, two code paths for one operation, recovery code only exercised when sfnt rejects) for a same-week fix. The right answer is the deeper one: drop sfnt entirely from the metric-reading path and own the parser end-to-end.

Scoping under follow-up. Closing this PR; the user-reported bug remains unfixed for one more week.

Architectural debt this PR would have shipped:
- `isUnsupportedCmapErr` substring match against sfnt's unexported error message
- `appendStubCmap` + `parseTTFWithStubbedCmap` recovery surgery (~75 LOC of stub-and-rewire that exists only to satisfy sfnt)
- `folioCmap != nil` branch in `GlyphIndex` creating two paths for the same operation
- Brittle joint at every `golang.org/x/image` bump

Phase 2 removes all of these by replacing sfnt's metric reads with direct parsers for `head`, `hhea`, `maxp`, `hmtx`, `OS/2`, `name`. Cmap parser from this PR carries forward unchanged. Estimated +300 LOC net, +1 week.

PR #262, opened 2026-05-01
font: drop sfnt dependency by parsing tables directly (#260)

### LINKED ISSUE #227: CJK fail render some Chinese Characters.

### LINKED ISSUE #248: font: large CJK fonts hit golang.org/x/image/sfnt's hardcoded maxCmapSegments=20000 limit

### LINKED ISSUE #260: font: drop sfnt dependency by parsing head/hhea/maxp/hmtx/OS-2/name from raw bytes (Phase 2 of #248)
Replaces the closed PR #251 (Phase 1, recovery-path approach) with the structurally cleaner answer to #248: own the entire metric-reading surface so `golang.org/x/image/font/sfnt`'s hardcoded limits stop being our problem.

Closes #227 (#235 fixed TTC dispatch but the user-reported fonts still trip sfnt's `maxCmapSegments=20000`). Closes #248 (the cmap-segments limit itself). Removes the heuristic seam Phase 1 would have shipped (substring-matching sfnt's error message, stub-and-rewire surgery, two code paths for `GlyphIndex`).

## Why Phase 2 instead of Phase 1

Phase 1 (PR #251, closed) routed around the sfnt cmap limit with a recovery path: detect sfnt's error by substring, parse cmap from raw bytes, install a 22-byte stub cmap so sfnt accepts the rest of the font. This works for the user but ships architectural debt:
- `isUnsupportedCmapErr` substring-matches sfnt's unexported error message
- ~75 LOC of `appendStubCmap` / `parseTTFWithStubbedCmap` exists only to satisfy sfnt
- `folioCmap != nil` branch in `GlyphIndex` creates two paths for one operation
- Every `golang.org/x/image` bump risks silently breaking the recovery on every CJK font

Phase 2 removes all of these by removing sfnt from the operations Folio uses for metrics. Cmap is the easy part; the rest is straightforward binary parsing of well-specified tables.

## What sfnt does for Folio today

7 distinct operations across 14 call sites in `font/truetype.go`:

| sfnt operation | Folio caller | Source table |
|---|---|---|
| `sfnt.Parse(data)` | ParseTTF | (validates structure) |
| `f.font.Name(buf, ID)` | PostScriptName, NameID | `name` |
| `f.font.UnitsPerEm()` | UnitsPerEm | `head` |
| `f.font.GlyphIndex(buf, r)` | GlyphIndex | `cmap` |
| `f.font.GlyphAdvance(buf, gid, ppem, hinting)` | GlyphAdvance | `hmtx` |
| `f.font.Metrics(buf, ppem, hinting)` | Ascent, Descent | `hhea` / `OS/2` |
| `f.font.Bounds(buf, ppem, hinting)` | BBox | `head` |
| `f.font.NumGlyphs()` | NumGlyphs | `maxp` |

Folio is already 60-70% of the way there: `parseTTFTables`, `parseLoca`, `findTable`, `extractTTCFont`, `ItalicAngle`, `CapHeight`, `Flags`, `StemV`, plus the cmap parser drafted in PR #251 — all parse from raw bytes today.

## Scope estimate

**~300 net new LOC** (~380 added, ~80 removed):

| Parser | LOC | Notes |
|---|---|---|
| `head` reader | 30 | UnitsPerEm, font BBox, indexToLocFormat |
| `hhea` reader | 25 | Ascent, Descent, LineGap, numHMetrics |
| `maxp` reader | 10 | numGlyphs |
| `hmtx` reader | 30 | Per-glyph advance + LSB; structure understood from existing `rebuildHmtx` |
| `OS/2` reader | 40 | Typo/win metrics; fsSelection USE_TYPO_METRICS bit selection is the subtle part |
| `name` reader | 80 | Unicode/Mac platform selection, UTF-16BE decoding — the trickiest piece |
| **Tests** | 250 | Synthetic-fixture per parser + cross-table consistency |

**Removed from PR #251's draft:**
- `appendStubCmap` (50 LOC)
- `isUnsupportedCmapErr`, `parseTTFWithStubbedCmap` (25 LOC)
- `folioCmap` field + branch in `GlyphIndex` (5 LOC)

**Carries forward unchanged from PR #251:**
- `font/cmap.go` (`parseCmapTable`, `parseCmapFormat4`, `parseCmapFormat12`, `scoreSubtable`)
- All cmap tests (synthetic format-4/format-12, subtable picking, malformed input)

## Out of scope

- **CFF/OTF outline support.** Folio's `Subset` only handles TrueType `glyf` today; CFF fonts parse via sfnt but fail at subset. Phase 2 maintains that scope. Adding minimal CFF reading is +200-400 LOC and a separate decision.
- **Hinting / TrueType program execution.** Viewers do hinting, not us.
- **Variable fonts** (`fvar`/`gvar`/`HVAR`).
- **Color font tables** (`COLR`/`CPAL`/`sbix`/`CBDT`).

## Risk profile

What goes away (vs PR #251):
- Substring matching against an unexported sfnt error
- Stub-and-rewire surgery
- Two code paths for `GlyphIndex`
- Bump-tracking dependency on `golang.org/x/image/font/sfnt`'s error wording

What we take ownership of:
- `name` table encoding selection (Mac vs Windows platform, language/script handling)
- `OS/2` metric selection (typo vs win metrics, USE_TYPO_METRICS bit interpretation)
- Cross-table validation (numGlyphs consistency across maxp / hmtx / loca)

Both are well-specified in ISO/IEC 14496-22 §5; sfnt's choices for the ambiguous parts are pragmatic rather than canonical, so we're not losing reference behavior.

## Acceptance criteria

- [ ] `font/truetype.go` no longer imports `golang.org/x/image/font/sfnt` from the metric-reading path. (Subset code can keep using `parseTTFTables` etc.)
- [ ] Each new parser has synthetic-fixture tests + a real-system test covering at least one TTF and one OTF candidate
- [ ] Cmap recovery path from PR #251 carries forward unchanged
- [ ] User-reported #227 scenario: STHeiti / msyh.ttc / NotoSansCJK-Regular.ttc render with correct text-extraction round-trip
- [ ] Existing test corpus passes byte-identically (no metric-rounding regressions in subset/embed output)
- [ ] Both `!js` and `js`/`wasm` builds clean
- [ ] gofmt + race-detector clean

## Estimated effort

~1 week for implementation + tests + review-cycle integration. Larger than PR #251's couple-of-days but eliminates the architectural debt rather than shipping it.

carlos7ags:
## Summary

- Replaces `golang.org/x/image/font/sfnt` in Folio's metric path with seven new in-tree parsers (one file per OFF table): `font/{head,hhea,maxp,hmtx,os2,name,cmap}.go`. `font/sfnt_tables.go` orchestrates table-directory parsing and per-table decoding into a `parsedFont` struct that backs every `Face` method on `sfntFace`.
- Closes #248: sfnt's hardcoded `maxCmapSegments = 20000` rejected Microsoft YaHei, Noto Sans CJK, and STHeiti — the most common CJK fonts in the wild. The new cmap parser bounds-checks each segment against the table's declared length rather than enforcing a hard cap; a 25,000-group format-12 synthetic fixture pins the regression.
- Fixes #227 (the user-reported CJK rendering failure caused by the upstream cmap limit). Smoke-tested end-to-end on macOS: `/System/Library/Fonts/STHeiti Light.ttc` loaded, HTML containing 中华人民共和国 converted to PDF, parsed back, `page.ExtractText()` round-trips the original text.
- Avoids the architectural debt of the closed PR #251, which recovered from sfnt's failure by substring-matching its unexported error message.

## Scope

- TrueType `glyf`/`loca` outline support is unchanged.
- CFF/OTF outline tables stay as opaque byte slices in `parsedFont.rawTables` for embedding. Folio's `Subset` already only handled TrueType outlines, so CFF fonts continue to load and embed but cannot be subset (#260).
- WOFF1 dispatch and TTC extraction are untouched — both still feed bytes through `ParseTTF`, which now goes through the in-tree parsers.

## Behaviour parity

- `Ascent` / `Descent` read OS/2 `sTypoAscender` / `sTypoDescender` (with `hhea` fallback for older TrueType lacking OS/2), matching what `sfnt.Metrics()` produced but without the `fixed.Int26_6` detour.
- `BBox` reads `head.{xMin,yMin,xMax,yMax}` directly. The `head` table already stores PDF Y-up coordinates, so the previous Y-axis flip (negating sfnt's Y-down output) is gone.
- `GlyphIndex` / `GlyphAdvance` read `parsedFont.cmap` and `parsedFont.advances` directly — no fixed-point conversion.
- `PostScriptName` resolves to NameID 6 with NameID 4 fallback. Encoding selection prefers Windows Platform 3 records, then Unicode-platform, then MacRoman.
- `GIDToUnicode` inverts the parsed cmap (lowest rune wins), matching the previous BMP-ascending scan in `BuildGIDToUnicode`.

## Safety

- Every byte read is preceded by a `uint64`-arithmetic bounds check against the table length. Malformed input returns a wrapped `ErrTruncated` or `ErrCorruptTable`; no parser can panic.
- Each parser file cites the relevant ISO/IEC 14496-22 §5 clause and lists field offsets in its top-of-function comment.

## Test plan

- [x] `go test ./...` passes locally on darwin/arm64 (Go 1.26).
- [x] `gofmt -s -l .` reports nothing.
- [x] `GOOS=js GOARCH=wasm go build` succeeds for `font`, `html`, `layout`, `core`, `content`, `reader`, `document`. (`export` has a preexisting wasm issue unrelated to this PR.)
- [x] New per-parser tests cover: synthetic-fixture round-trip, truncated input → `errors.Is(err, ErrTruncated)`, corrupt-magic / corrupt-counts → `errors.Is(err, ErrCorruptTable)`, opportunistic real-system-font validation.
- [x] `font/cmap_test.go::TestParseCmapFormat12LargeGroupCount` proves the parser handles 25k groups (above sfnt's old 20k cap). Regression-pin verified by stubbing the function to return empty — the test fails, confirming load-bearing behaviour.
- [x] `font/cmap_test.go::TestParseCmapSTHeitiOpportunistic` validates the user-reported font on darwin: `face.GlyphIndex('中')` returns non-zero.
- [x] End-to-end STHeiti smoke (deleted before commit): HTML → PDF → reader → `ExtractText` round-trips the original CJK text.

Closes #260, #248, #227.


carlos7ags:
Addressed feedback from both review passes in 7a77e66 (rebased on `origin/main` first to clear the unrelated #261 regression).

**Code review (1 substantive fix):**
- Pre-fix `Ascent`/`Descent` used `os2.sTypoAscender`/`sTypoDescender` unconditionally when OS/2 was present, but sfnt always uses hhea (`grep "fsSelection|sTypoAscender|UseTypoMetrics"` in sfnt v0.39.0 returns zero matches). The "matches sfnt" claim was wrong — for any font where hhea differs from typo metrics, PDF FontDescriptor numbers shifted between v0.7.1 and this PR. Wired the existing-but-unused `useTypoMetrics()` method into `Ascent`/`Descent`: USE_TYPO_METRICS bit SET → typo metrics; UNSET → hhea (sfnt's behavior). Preserves byte-identical output for every font the prior code shipped to. CHANGELOG updated to describe the new selection logic accurately.

**Test review (1 blocker, 2 must-haves):**
- **Blocker (rebase)**: branch was one commit behind `origin/main` and silently reverted #261's `column-rule` fix. `git rebase origin/main` cleared the unrelated `html/` deletions; `git diff origin/main..HEAD -- html/` is now empty.
- **Must-have**: `integration/cjk_drop_sfnt_roundtrip_test.go` — pins that this PR actually fixes #227 end-to-end. Renders Chinese via `@font-face url(STHeiti)` with `BaseFS=nil`, parses the PDF, calls `page.ExtractText()`, asserts byte-perfect round-trip through the embedded subset's `/ToUnicode` CMap. The agent ran this as a discarded smoke; promoting to a real test pins the most important pre-merge correctness claim.
- **Must-have**: `TestParseCmapFormat4IdRangeOffset` — exercises the format-4 indirect-offset path (`idRangeOffset != 0`) in `parseCmapFormat4`. The "obscure indexing trick" pointer arithmetic at `cmap.go:184` is non-trivial and was previously only covered by real fonts loaded opportunistically. Verified the test fails when the `readOff` calculation is misaligned ("rune U+0040: got GID 0, want 100"); restored.

**Deferred per both reviewers' explicit scope guidance:**
- CFF/OTF parse test (Folio's subset doesn't support CFF; dispatch already covered by `TestParseFontDispatchSurface`)
- Deepening STHeiti opportunistic test to multiple codepoints
- Stub-verifying the other 6 parsers
- Removing the unused `rawTables` method (cosmetic)

Full `go test ./...` green; `gofmt -s -l .` empty; both `!js` and `js`/`wasm` builds clean.

Issue #299, opened 2026-05-14
font: CFF subsetting for CID-keyed CJK fonts (#295)


carlos7ags:
## Problem

Fixes the cause behind #295 (14 MB PDFs from a single CJK character on
Ubuntu).

`NotoSansCJK-Regular.ttc` is a TrueType Collection whose faces are
CFF-flavored OpenType (sfntVersion `OTTO`, glyph data in the `CFF `
table). Folio's subsetter at `font/subset.go` only supports TrueType
outlines (`glyf` / `loca`). When `Subset` is called on a CFF face, it
fails with `ErrMissingTable`, and `font/embed.go:167` falls back to
embedding the full font (~14 MB).

The auto-fallback in `html/converter.go:717` selects NotoSansCJK first
on Linux, so any HTML containing a CJK glyph triggers this path. The
same fallback picks `msyh.ttc` on Windows, which has TrueType outlines
and subsets correctly — hence the OS asymmetry the reporter observed.

## Approach

Add a CFF v1 subsetter alongside the existing TrueType one and branch
the embed path on the source font's outline format. CFF2 (variable
fonts) is out of scope; NotoSansCJK and other CID-keyed CJK fonts are
CFF v1.

## Phase 1 — embed path dispatch

- Add `Face.IsCFF() bool` (package-private check on `rawTables["CFF "]`).
- Add `face.CFFData() []byte` accessor returning the raw `CFF ` table
  bytes (the PDF `FontFile3 /CIDFontType0C` stream expects raw CFF,
  not the sfnt wrapper).
- New `buildObjectsCFF` mirroring `BuildObjects` but emitting:
  - `/FontFile3` (not `/FontFile2`) with `/Subtype /CIDFontType0C`,
    no `/Length1`.
  - `/Subtype /CIDFontType0` on the descendant CIDFont.
  - No `/CIDToGIDMap` — implicit Identity in CFF.
- Existing `BuildObjects` keeps the TrueType path; the dispatch lives
  at the top.

`/W`, ToUnicode CMap, Type0 wrapper, `Identity-H`, `Adobe-Identity-0`
ordering all reused unchanged.

## Phase 2 — CFF parser (read-only)

Parser for Adobe TN #5176 §§5–18 covering Header, Name INDEX, Top DICT
INDEX, String INDEX, Global Subr INDEX, Charsets, FDSelect (format 0
and 3), CharStrings INDEX, FDArray (per-FD font DICTs), Private DICTs,
[...]

PR #300, opened 2026-05-15
font: CFF subsetting for CID-keyed CJK fonts (closes #295)

### LINKED ISSUE #295: Convert HTML to PDF, got large file size with Chinese character insert in Ubuntu
**Describe the bug**
In Ubuntu:
1. Insert several Chinese character in Header Title of examples/html-to-pdf/main.go, output pdf size rapidly expanding from 12K(11314) to 14M(13791542)
2. Mask NotoSansCJK in examples/fonts/main.go, output pdf size reduce from 14M(13806895) to 24K(23773)

- It seems that folio import whole fonts for Chinese or CJK in Ubuntu.
- What's the default CJK/Chinese font used in Linux?
- Is it caused by the difference in the ttc and ttf formats？

**To reproduce**

### examples/html-to-pdf
```
cd examples/html-to-pdf

# orginal
git checkout main.go
git diff main.go
go run main.go
ls -l
# total 24
# -rw-rw-r-- 1 wangyaoshen wangyaoshen 12287 May  7 10:00 main.go
# -rw-rw-r-- 1 wangyaoshen wangyaoshen 11314 May  7 10:00 report.pdf

# insert Chinese Character
sed 's|<h1>Quarterly Report</h1>|<h1>Quarterly 报告</h1>|' -i main.go
# git diff main.go
# diff --git a/examples/html-to-pdf/main.go b/examples/html-to-pdf/main.go
# index 5e55f78..15080db 100644
# --- a/examples/html-to-pdf/main.go
# +++ b/examples/html-to-pdf/main.go
# @@ -120,7 +120,7 @@ const reportHTML = `<!DOCTYPE html>
#  <body>
#    <!-- ============ PAGE 1: Dashboard ============ -->
#    <div class="header-band">
# -    <h1>Quarterly Report</h1>
# +    <h1>Quarterly 报告</h1>
#      <div class="sub">Q4 2026 &mdash; Apex Capital Partners &mdash; Confidential</div>
#    </div>
go run main.go
#    ls -l
#    total 13484
#    -rw-rw-r-- 1 wangyaoshen wangyaoshen    12287 May  7 10:05 main.go
#    -rw-rw-r-- 1 wangyaoshen wangyaoshen 13791542 May  7 10:05 report.pdf
```

### examples/fonts
```
cd examples/fonts/

# orginal
git checkout main.go
git diff main.go
go run main.go
ls -l
total 13492
-rw-rw-r-- 1 wangyaoshen wangyaoshen 13806895 May  7 10:23 fonts.pdf
-rw-rw-r-- 1 wangyaoshen wangyaoshen     5306 May  7 10:22 main.go

# mask NotoSansCJK 
sed 's|{"NotoSansCJK"|//{"NotoSansCJK"|' -i main.go
git diff main.go
# diff --git a/examples/fonts/main.go b/examples/fonts/main.go
# index 2542ac1..ebedc5f 100644
# --- a/examples/fonts/main.go
# +++ b/examples/fonts/main.go
# @@ -130,7 +130,7 @@ func discoverFonts() []fontEntry {
#                 candidates = []fontEntry{
#                         {"NotoSans", "Noto Sans", "/usr/share/fonts/truetype/noto/NotoSans-Regular.ttf", false},
#                         {"DejaVu", "DejaVu Sans", "/usr/share/fonts/truetype/dejavu/DejaVuSans.ttf", false},
# -                       {"NotoSansCJK", "Noto Sans CJK", "/usr/share/fonts/opentype/noto/NotoSansCJK-Regular.ttc", true},
# +                       //{"NotoSansCJK", "Noto Sans CJK", "/usr/share/fonts/opentype/noto/NotoSansCJK-Regular.ttc", true},
#                 }
#         case "windows":
#                 candidates = []fontEntry{
go run main.go
ls -l
total 32
-rw-rw-r-- 1 wangyaoshen wangyaoshen 23773 May  7 10:30 fonts.pdf
-rw-rw-r-- 1 wangyaoshen wangyaoshen  5308 May  7 10:30 main.go
```

**Expected behavior**
PDF file size expand slower with CJK font in Linux.

**Environment**
- Go version: go version go1.26.0 linux/amd64
- OS: Linux ubuntu 5.15.0-163-generic # 173-Ubuntu SMP Tue Oct 14 17:51:00 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
- Folio version/commit: v0.7.1-135-g0fd5b02

**Additional context**

|source|key words|size in Windows|size in Ubuntu|
|-------|-----------|------------------|----------------|
|html-to-pdf|Quarterly Report|11314|11314|
|html-to-pdf|Quarterly 报告|15706|13791542|
|html-to-pdf|季度 Report|15907|13791535|
|html-to-pdf|季度 报告|16953|13791575|

carlos7ags:
## Summary

Adds a CFF v1 subsetter for CID-keyed fonts and routes CFF-flavored faces through a spec-compliant PDF object graph. Closes #295 and #299.

The reporter's exact reproducer — `examples/html-to-pdf` with a single Chinese character on a system whose CJK auto-fallback picks NotoSansCJK (Linux) or Hiragino Sans GB (macOS) — drops from **14 MB to 16 KB**. The embed-path stream payload for the same fixture is **11.1 MB → 173 KB (1.6%)** before PDF FlateDecode further compresses it.

## What changed

Six commits, each independently reviewable:

| Commit | Phase | What |
|---|---|---|
| 1 | Dispatch + PDF object graph | `EmbeddedFont.BuildObjects` branches on CFF table presence; CFF faces route through a new builder that emits `/FontFile3` + `/CIDFontType0C` stream + `/CIDFontType0` descendant + no `/CIDToGIDMap` (TrueType path untouched) |
| 2 | CID-keyed gate | Adds a defensive Top-DICT walker (`isCIDKeyedCFFv1`) and excludes non-CID CFF + CFF2 from the new path. Tightens test assertions via `*core.PdfDictionary.Get()` casts |
| 3 | Read-only parser | `parseCFF` decodes INDEX / DICT / Charset / FDSelect / FDArray with per-operand byte spans for downstream byte-level rewriting |
| 4 | Parser hardening | Per-operand byte ranges in `cffDictEntry`, FDSelect format-3 sentinel check, `FuzzParseCFF` / `FuzzParseCFFDict`, complete synthetic CID-keyed CFF builder |
| 5 | Subsetter (closes #295) | Type 2 charstring reachability walker + three-pass byte-level assembly with fixed 5-byte longint operand encoding. Wires `SubsetCFF` into the embed path with the standard `XXXXXX+` subset tag prefix |
| 6 | Review fixes | Rejects predefined-charset values for CID-keyed CFF, distinguishable-charstring + gsubr-pruning + non-longint-operand blind-spot tests, FDSelect format-3 coverage, bounded-work walker assertion |

## Algorithm — Phase 3 in one paragraph

The subsetter rebuilds CharStrings INDEX with unkept GIDs replaced by a single `endchar` (0x0E) and the Global + per-FD Local Subr INDEXes with subroutines that aren't transitively reached replaced by a single `return` (0x0B). A partial Type 2 interpreter walks the operand stack precisely enough to read every `callsubr` / `callgsubr` argument and the stem count consumed by `hintmask` / `cntrmask`, with a pessimistic fallback when a call's argument was supplied by a caller across a frame boundary. Every offset operand in the rebuilt Top DICT, FDArray font dicts, and per-FD Private DICTs uses fixed 5-byte longint encoding so section sizes are independent of offset values, which means layout is a single forward pass: build sections with zeroed placeholders, sum section lengths, patch placeholders in place, concatenate. Header, Name INDEX, String INDEX, Charset, FDSelect are copied verbatim from source.

## Test plan

- [x] `go test ./...` green
- [x] `FuzzParseCFF`, `FuzzParseCFFDict`, `FuzzSubsetCFF` clean at >1M execs each (5-10s budget locally)
- [x] Real-font subset regression: `TestSubsetCFFRealFontSize` asserts <10% of source on Hiragino Sans GB.ttc (actual: 1.6%)
- [x] Embed-path regression: `TestBuildObjectsCFFSubsetsFontStream` asserts the same against the stream payload through `EmbeddedFont.BuildObjects` (actual: 1.6%)
- [x] Issue #295 end-to-end reproducer verified locally: `examples/html-to-pdf/main.go` with `<h1>Quarterly 报告</h1>` produces a 16 KB PDF (was 14 MB pre-fix on Ubuntu)
- [x] Rebased on current `origin/main`

## Notes

- **No public API changes.** `Face` interface unchanged; CFF detection is internal via package-private `cffFace`.
- **Fallback policy.** On `SubsetCFF` failure, `buildObjectsCFF` falls back to embedding the source CFF unchanged — same silent-fallback semantics as the TrueType path at `font/embed.go:167`. The fallback is now defensive: every structural failure mode in `parseCFF` has a sentinel-wrapped error.
- **Non-CID CFF.** Plain Latin `.otf` fonts (e.g. STIXGeneral) are correctly excluded from the new path; they need a `/Type1C` + non-composite font dict path that this PR does not add. The legacy embed path handles them as before.
- **CFF2.** Variable-CFF fonts are detected and routed back to the legacy path; a CFF2 embed path can land separately when there's user demand.
- **Out of scope.** Full subroutine re-indexing (would shrink ~173 KB to ~30 KB). The current 1.6%-of-source result is already adequate to close #295; the additional shrink is a separate work item.

### DIFF
--- a/document/document.go
+++ b/document/document.go
@@ -620,6 +620,17 @@ func (d *Document) WriteToWithOptions(w io.Writer, opts WriteOptions) (int64, er
 	pageRefs := make([]*core.PdfIndirectReference, 0, len(allPages))
 	pageDicts := make([]*core.PdfDictionary, 0, len(allPages))
 
+	// Document-wide font caches. A given *font.EmbeddedFont (or
+	// *font.Standard) is typically registered on every page that
+	// uses it, so without these caches BuildObjects / Dict would run
+	// once per page and the writer would emit a fresh /FontFile3 +
+	// CIDFont + descriptor for every page. For a 20-page document
+	// referencing one CJK font, that's 20× the embedded font data.
+	// The cache shares the Type0 indirect reference across pages so
+	// the font program is embedded exactly once per document.
+	embeddedFontRefs := make(map[*font.EmbeddedFont]*core.PdfIndirectReference)
+	standardFontRefs := make(map[*font.Standard]*core.PdfIndirectReference)
+
 	// First pass: build page objects.
 	for pageIdx, page := range allPages {
 		pageDict := core.NewPdfDictionary()
@@ -658,12 +669,20 @@ func (d *Document) WriteToWithOptions(w io.Writer, opts WriteOptions) (int64, er
 			fontDict := core.NewPdfDictionary()
 			for _, entry := range page.fonts {
 				if entry.standard != nil {
-					fontObjRef := writer.AddObject(entry.standard.Dict())
-					fontDict.Set(entry.name, fontObjRef)
+					ref, ok := standardFontRefs[entry.standard]
+					if !ok {
+						ref = writer.AddObject(entry.standard.Dict())
+						standardFontRefs[entry.standard] = ref
+					}
+					fontDict.Set(entry.name, ref)
 				} else if entry.embedded != nil {
-					type0 := entry.embedded.BuildObjects(writer.AddObject)
-					fontObjRef := writer.AddObject(type0)
-					fontDict.Set(entry.name, fontObjRef)
+					ref, ok := embeddedFontRefs[entry.embedded]
+					if !ok {
+						type0 := entry.embedded.BuildObjects(writer.AddObject)
+						ref = writer.AddObject(type0)
+						embeddedFontRefs[entry.embedded] = ref
+					}
+					fontDict.Set(entry.name, ref)
 				}
 			}
 			resources.Set("Font", fontDict)
--- a/document/document_test.go
+++ b/document/document_test.go
@@ -165,6 +165,53 @@ func TestAddTextMultiplePages(t *testing.T) {
 	}
 }
 
+// TestFontDedupAcrossPages pins the document-level font-sharing fix:
+// the same *font.EmbeddedFont (and *font.Standard) referenced by
+// multiple pages must register exactly one font object in the PDF,
+// not one per page. Pre-fix, a 20-page document with one shared CJK
+// embedded font produced 20 /FontFile2 streams; post-fix it produces
+// one. The reporter of issue #295 specifically noticed this — "each
+// page contains one copy of CJK fonts" — and the multi-MB blowup
+// from per-page embedding was the visible symptom.
+func TestFontDedupAcrossPages(t *testing.T) {
+	face, err := font.LoadTTF(testFontPath(t))
+	if err != nil {
+		t.Fatalf("LoadTTF: %v", err)
+	}
+
+	doc := NewDocument(PageSizeLetter)
+	ef := font.NewEmbeddedFont(face)
+	for i := range 5 {
+		p := doc.AddPage()
+		// Both standard (Helvetica) and embedded fonts go on every
+		// page so we test both dedup paths in one assertion.
+		p.AddText("Std", font.Helvetica, 12, 72, 720)
+		_ = i
+		p.AddTextEmbedded("a", ef, 12, 72, 700)
+	}
+
+	var buf bytes.Buffer
+	if _, err := doc.WriteTo(&buf); err != nil {
+		t.Fatalf("WriteTo: %v", err)
+	}
+	pdf := buf.String()
+
+	// The embedded TrueType path emits exactly one /FontFile2
+	// reference per unique embedded font; with five pages sharing
+	// one EmbeddedFont we want exactly one.
+	if got := strings.Count(pdf, "/FontFile2"); got != 1 {
+		t.Errorf("/FontFile2 occurrences = %d, want 1 (per-document share)", got)
+	}
+	// Helvetica is a Standard font: its dict is small but should
+	// still be emitted once, not five times.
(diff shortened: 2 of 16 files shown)


carlos7ags:
Follow-up after re-reading the reporter's comment on #295 ("each page contains one copy of CJK fonts").

Verified end-to-end that the reporter's exact 2-page reproducer still embedded the font twice with the CFF subsetter alone. Root cause was at the document writer (not the CFF path): \`EmbeddedFont.BuildObjects\` and \`Standard.Dict()\` were called once per page per font instead of once per document.

Added commit \`61dc047\` that caches the indirect reference per \`*font.EmbeddedFont\` and \`*font.Standard\` across the page-write loop. Applies to both /FontFile3 (CFF) and /FontFile2 (TrueType) paths.

Reporter's exact 2-page reproducer:
- Original: ~27 MB
- After CFF subsetter only: ~350 KB
- After this commit: ~5 KB

20-page synthetic test: 94,777 → 10,917 bytes; structural inspection shows one /FontFile2 and one /CIDFontType2 (was twenty of each).

\`TestFontDedupAcrossPages\` pins the new behavior.

Issue #358, opened 2026-06-09
html/layout: nested list markers stack at the container left margin instead of indenting per level


carlos7ags:
**Describe the bug**

Nested ordered/unordered list markers are all drawn at the container's left margin, regardless of nesting depth — only the item *body* text indents per level. A reader of a multi-level (e.g. legal) document expects each level's number to indent under its parent's text:

```
1.2.  "Services" means ...
      1.2.1.  Each statement of work ...
```

but Folio renders the sub-number near the same left column as its parent:

```
1.2.  "Services" means ...
1.2.1.        Each statement of work ...
```

**Evidence (pre-existing, independent of #356)**

A plain three-level ordered list (`PlanLayout`, draw x-coordinates of each marker/body token) shows every marker at x=0 while bodies cascade right:

```
x=0.0  "1."          (level 1 marker)
x=49.4 "one ..."     (level 1 body)
x=0.0  "1."          (level 2 marker)   <- not indented
x=89.2 "item"        (level 2 body)
x=0.0  "1."          (level 3 marker)   <- not indented
x=116  "item"        (level 3 body)
```

The marker is drawn left-aligned at `absX` across the full `totalIndent` in `layout/list.go` (`planAt`/`layoutAt`), so deeper levels never shift the marker, only the text column.

**Expected**

Each nested level's marker should indent so the number aligns with (or just left of) its parent item's content edge, matching how browsers render nested ordered lists.

**Notes**

Surfaced while implementing multi-level numbering (#356); that feature produces the correct ordinals within the current layout model. Fixing marker indentation is separate core-layout work in `layout/list.go` and is tracked here.

Issue #356, opened 2026-06-07
html/layout: Multi-level (legal) numbering for nested ordered lists + hanging marker (list-style-position: outside)


daniils-keyos:
**Describe the feature**
Two related improvements to ordered-list (`<ol>`/`<li>`) rendering:

1. Multi-level ("legal") numbering for nested ordered lists. When `<ol>`s are nested, sub-items should be able to render hierarchical, dotted ordinals that include the ancestors'  numbers — 1, 1.1, 1.1.1, 1.2, 2, … — instead of each nested list restarting at 1. Ideally exposed the CSS way via counters, e.g.:
```
ol { counter-reset: item; list-style: none; }
li { counter-increment: item; }
li::marker { content: counters(item, ".") ". "; }
```

i.e. support for counter-reset / counter-increment on lists/items and the counters(name, sep) function in `::marker { content }`. (A built-in list-style-type shorthand for nested decimal would also work, but counters() is the general primitive.)
 
2. Hanging indent for the list marker (list-style-position: outside). The marker should sit in a left gutter outside the item's content box, and the item's content — the title, any following block paragraphs, and wrapped continuation lines — should all align at one left edge to the right of the marker. Today the marker is laid out inline with the first line, so wrapped lines and block children align under the marker rather than under the text.
desired:
```
7.1  The term ("Term") of this Agreement shall be
     as set forth below, unless sooner terminated as provided ...
```
today:
```
7.1 The term ("Term") of this Agreement shall be
as set forth below, unless sooner terminated as provided ...
````
**Use case**
We render legal agreements to PDF from stored HTML. These documents are structured as deeply nested ordered lists where every clause is referenced by its full path so flat per-level numbering (1, 1, 1 at each depth) is not just cosmetic — it changes the meaning. Multi-level numbering is the standard convention for contracts, statutes, specs, and technical documentation.

The hanging indent matters for readability of those same clauses: a clause is typically `<li><u>Title.</u><div>…long body…</div><ol>…sub-clauses…</ol></li>`, and without an outside marker the wrapped body text and the sub-list slide back under the number, making the hierarchy hard to scan. With list-style-position: outside, the number hangs in the gutter and the whole clause body (title + description + nested items) shares a clean left margin.

[...]

PR #363, opened 2026-06-09
html/layout: optional flush (outline) list-marker alignment

carlos7ags:
Follow-up to #358/#360. Adds an opt-in mode that draws every nested list
marker in the container's left column (all numbers share one left edge) while
the body text still cascades per level — the "outline" style common in legal
documents — as an alternative to the per-level marker indent that is now the
default.

### Comparison
Same Master Services Agreement (`examples/legal-numbering`), rendered both ways:

```
Nested (default)                 Flush (FlushListMarkers: true)
1.   Definitions.                1.   Definitions.
  1.1.  "Agreement" ...            1.1.  "Agreement" ...
  1.2.  "Services" ...             1.2.  "Services" ...
    1.2.1.  Each statement ...       1.2.1.    Each statement ...
    1.2.2.  In the event ...         1.2.2.    In the event ...
2.   Term and Termination.       2.   Term and Termination.
```

In nested, `1.2.1.`/`1.2.2.` indent under "Services"; in flush they stay in the
left column and the clause bodies hang. Reproduce locally by rendering the
`legal-numbering` example HTML with `&fhtml.Options{FlushListMarkers: true}`.

### API
Document-wide, off by default:

```go
doc.AddHTML(html, &fhtml.Options{FlushListMarkers: true})
```

Underneath, `layout.List.SetMarkerFlush(bool)` toggles it; sub-lists inherit
the value. Only the left-to-right outside path is affected; RTL and
`list-style-position: inside` are unchanged. Default (nested) output is
byte-identical.

### Scope
Kept intentionally minimal: a single document-wide option, no new CSS
property. If per-`<ol>` control within one document is needed later, a
(non-standard) CSS property can layer on top of the same `SetMarkerFlush`
plumbing.

### Tests
`TestNestedMarkerFlush` (layout: all markers share x=0 while bodies cascade) and
`TestFlushListMarkersOption` (the Options toggle reaches the layout and changes
the render while numbering stays intact). Full module suite passes.

### DIFF
--- a/html/converter_list.go
+++ b/html/converter_list.go
@@ -103,6 +103,13 @@ func (c *converter) convertList(n *html.Node, style computedStyle, ordered bool)
 		list.SetMarkerInside(true)
 	}
 
+	// FlushListMarkers (document-wide option) draws all nested markers in one
+	// left column instead of indenting per level. Sub-lists inherit it via the
+	// AddItem*WithSubList constructors.
+	if c.opts.FlushListMarkers {
+		list.SetMarkerFlush(true)
+	}
+
 	c.populateList(n, list, style)
 
 	return []layout.Element{list}
--- a/html/flush_markers_test.go
+++ b/html/flush_markers_test.go
@@ -1,0 +1,47 @@
+// Copyright 2026 Carlos Munoz and the Folio Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package html
+
+import (
+	"strings"
+	"testing"
+)
+
+// TestFlushListMarkersOption verifies the document-wide Options.FlushListMarkers
+// toggle reaches the layout: it changes how a nested ordered list renders while
+// leaving the numbering intact. The precise flush geometry (all markers in one
+// left column) is asserted at the layout level in TestNestedMarkerFlush; here
+// we only confirm the option is wired through Convert.
+func TestFlushListMarkersOption(t *testing.T) {
+	const doc = `<ol><li>One<ol><li>Inner</li></ol></li><li>Two</li></ol>`
+
+	nested, err := Convert(doc, &Options{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	flush, err := Convert(doc, &Options{FlushListMarkers: true})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	nStream := renderStreamText(t, nested)
+	fStream := renderStreamText(t, flush)
+
+	// Numbering renders in both modes (top-level 1./2., nested 1.).
+	for _, want := range []string{"(1.) Tj", "(2.) Tj"} {
+		if !strings.Contains(nStream, want) {
+			t.Errorf("default render missing marker %q", want)
+		}
+		if !strings.Contains(fStream, want) {
+			t.Errorf("flush render missing marker %q", want)
+		}
+	}
+
+	// The option must change the layout (the nested marker moves from its
+	// indented column to the container's left column); otherwise it is not
+	// wired through Options.
+	if nStream == fStream {
+		t.Error("FlushListMarkers had no effect on the rendered stream — option not wired")
+	}
+}
--- a/layout/list.go
+++ b/layout/list.go
@@ -35,6 +35,7 @@ type List struct {
 	markerColor    *Color    // optional override color for markers
 	markerFontSize float64   // optional override font size for markers (0 = use list fontSize)
 	markerInside   bool      // CSS list-style-position: inside (marker flows inline)
+	markerFlush    bool      // draw every nested marker in the container's left column instead of indenting per level
 
 	// start is the ordinal offset for marker numbering. The marker for the
 	// item at slice index i is numbered (i + 1 + start). It defaults to 0 so
@@ -184,6 +185,18 @@ func (l *List) SetMarkerInside(inside bool) *List {
 	return l
 }
 
+// SetMarkerFlush controls how nested-list markers align. When flush is true,
+// every level's marker is drawn in the container's left column (all numbers
+// share one left edge) while the body text still cascades per level — the
+// "outline" style common in legal documents. The default (false) indents each
+// level's marker under its parent's content. Applies to the left-to-right
+// outside path; it is inherited by sub-lists. Sub-lists created later inherit
+// the value set on the parent at creation time.
+func (l *List) SetMarkerFlush(flush bool) *List {
+	l.markerFlush = flush
+	return l
+}
+
 // AddItem adds a text item to the list.
 func (l *List) AddItem(text string) *List {
 	l.items = append(l.items, listItem{text: normalizeText(text)})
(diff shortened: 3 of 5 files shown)


kuadrosx (approved):

html/converter_list.go
```diff
@@ -103,6 +103,13 @@ func (c *converter) convertList(n *html.Node, style computedStyle, ordered bool)
 		list.SetMarkerInside(true)
 	}
 
+	// FlushListMarkers (document-wide option) draws all nested markers in one
+	// left column instead of indenting per level. Sub-lists inherit it via the
+	// AddItem*WithSubList constructors.
+	if c.opts.FlushListMarkers {
```
kuadrosx, line 109: `capturedBase = 0` at `layout/list.go:729` and `:942` means "this `List`'s own left edge". That coincides with the container's left column only when the list was reached through `planAt`'s sub-list recursion, i.e. built via `AddItem*WithSubList`.

When an `<li>` takes the element path — a block child (`<p>`, `<br>`, `display: block`) or any box-model styling — `addElementListItem` wraps the item content in a `Div`, so the nested `<ol>` becomes an independent `layout.List` inside it. `planElementItem` then shifts that subtree via `offsetBlocks(plan.Blocks, totalIndent, curY)` (`layout/list.go:924`), so the child list's local zero already sits at the parent's content column. Zeroing `capturedBase` there is a no-op relative to the default.

Marker x on the unmodified `examples/legal-numbering` HTML (letter, 72pt margins):

| marker | flush=false | flush=true | |
|---|---|---|---|
| `1.` | 72 | 72 | |
| `1.1.` | 90 | 90 | unchanged |
| `1.2.` | 90 | 90 | unchanged |
| `1.2.1.` | 108 | 90 | one level only |
| `2.1.` | 90 | 90 | unchanged |

Every top-level `<li>` in that example contains a `<p>`, so all of them take the element path. The result is a two-column outline, not "all numbers share one left edge". A three-level all-inline `<ol>` does put every marker at x=72, so the mechanism is right — it just doesn't survive an element boundary.

Making the outline style work needs the accumulated outer offset threaded into element-path child lists (into `convertList`/`planAt`) rather than hard-zeroing `capturedBase`. Landing it as-is with the doc comments narrowed to "flush resets at items containing block content" is also defensible — but then the legal use case is only half-served, and the description's comparison table and its "reproduce locally by rendering the legal-numbering example" instruction need rewriting to match.

layout/list.go
```diff
@@ -209,12 +222,13 @@ func (l *List) AddItemRuns(runs []TextRun) *List {
...
-		fontSize: l.fontSize,
-		indent:   l.indent,
-		leading:  l.leading,
+		style:       ListUnordered,
+		font:        l.font,
+		embedded:    l.embedded,
+		fontSize:    l.fontSize,
+		indent:      l.indent,
+		leading:     l.leading,
+		markerFlush: l.markerFlush,
```
kuadrosx, line 231: This adds `markerFlush` to the inherited set, but `direction` and `markerInside` still aren't copied here (nor in the sibling constructors at `:250` and `:266`). Through the HTML fast path a nested `<ol>` never re-enters `convertList`, so a sub-list of an RTL or `inside` list is created LTR-outside — which puts it in exactly the `default:` draw branch flush now modifies.

So the option changes both, contrary to "RTL and `list-style-position: inside` are unchanged":

```
<ol dir="rtl"><li>One<ol><li>Inner</li></ol></li></ol>
  nested marker x: 90 (flush off) -> 72 (flush on)

<style>ol{list-style-position:inside}</style>
<ol><li>One<ol><li>Inner</li></ol></li></ol>
  nested marker x: 90 (flush off) -> 72 (flush on)
```

The RTL and inside draw branches do correctly ignore `capturedBase`, so the regression comes purely from the inheritance asymmetry. Two options: inherit `direction`/`markerInside` alongside `markerFlush` (fixes a latent nested-RTL bug too, but widens the diff), or gate the override on `l.direction != DirectionRTL && !l.markerInside` at the two `capturedBase` sites.

html/flush_markers_test.go
```diff
@@ -0,0 +1,47 @@
...
+		}
+		if !strings.Contains(fStream, want) {
+			t.Errorf("flush render missing marker %q", want)
+		}
+	}
+
+	// The option must change the layout (the nested marker moves from its
+	// indented column to the container's left column); otherwise it is not
+	// wired through Options.
+	if nStream == fStream {
```
kuadrosx, line 44: `nStream != fStream` can't distinguish "flush applied correctly" from "flush applied at the wrong nesting level" — it passes today despite the under-correction noted on `converter_list.go`, and would pass if the marker moved anywhere at all.

Asserting concrete marker x-coordinates, the way `TestNestedMarkerFlush` does at the layout level, over a three-level list with at least one `<li>` carrying a `<p>` child, would have caught it.

layout/list.go
```diff
@@ -184,6 +185,18 @@ func (l *List) SetMarkerInside(inside bool) *List {
...
 }
 
+// SetMarkerFlush controls how nested-list markers align. When flush is true,
+// every level's marker is drawn in the container's left column (all numbers
+// share one left edge) while the body text still cascades per level — the
+// "outline" style common in legal documents. The default (false) indents each
+// level's marker under its parent's content. Applies to the left-to-right
+// outside path; it is inherited by sub-lists. Sub-lists created later inherit
+// the value set on the parent at creation time.
+func (l *List) SetMarkerFlush(flush bool) *List {
```
kuadrosx, line 195: Worth making order-independent. The doc comment is honest about it and the converter happens to call this before `populateList`, but a layout-API caller who builds the tree first and configures after:

```go
l := NewList(font.Helvetica, 12).SetStyle(ListOrdered)
a := l.AddItemWithSubList("Level one")
a.AddItem("Level two")
l.SetMarkerFlush(true) // level 1 only, silently
```

gets flush on the top level only, with no error. Walking `l.items[i].subList` recursively in the setter would remove the footgun. Non-blocking.

kuadrosx (changes requested):
The plumbing here is clean and the default path is genuinely untouched, but I don't think the feature does what the description says yet. Two behavioural issues, both reproducible on the example the PR points at:

1. On element-path `<li>`s — any item with a block child or box-model styling — flush only sheds one level of indent instead of reaching the container's left column. That covers every top-level `<li>` in `examples/legal-numbering`, so the comparison table in the description doesn't match what the code renders.
2. The new `markerFlush` inheritance, without matching `direction` / `markerInside` inheritance, makes the option change RTL and `list-style-position: inside` documents — the opposite of the stated scope.

Details inline. Two smaller notes on the test's assertion strength and setter order-dependence, also inline.

One thing not anchorable to the diff: **no `CHANGELOG.md` entry.** `## Unreleased` is currently empty and this PR adds two exported surfaces (`html.Options.FlushListMarkers`, `layout.List.SetMarkerFlush`). The sibling `SetMarkerInside` from #356 is documented in the 0.9.1 `### Added` block, so this looks like a miss rather than a deliberate omission.

Verified on bd3917d: `gofmt`, `go vet`, and `go test ./layout ./html` are clean.

PR #375, opened 2026-06-29
Fix possible infinite loop in reader.collectPages

tgulacsi:
## Description

Handle cycles in pagetree walk.

## Problem
I've met with infinite loops in signing and page number retrieval - sorry, I don't have examples yet.

## Checklist

- [x] Code is formatted (`gofmt -s -w .`)
- [x] Tests pass (`make test`)
- [x] No references to external PDF libraries (clean room)


tgulacsi:
Test with 

    go test -run=Corrupt -v -timeout=30s
    
It should not timeout (but will on main).

carlos7ags (dismissed):

tgulacsi:
Tokenizer now advances all time, but this may be not the best solution.

PR #422, opened 2026-07-11
html: register element ids as PDF named destinations

kuadrosx:
## Problem

`<a href="#id">` links render as clickable annotations but navigate nowhere: an element's `id` is consumed only for CSS selector matching, so nothing registers it as a PDF named destination. The link emits a dangling `/GoTo (id)` string action pointing at a destination that is never defined — the generated PDF has an empty `/Dests`.

This is the behavior the `## [0.8.0]` CHANGELOG entry (#223) describes as fixed ("`layout.Anchor` element + auto-registered PDF named destinations"), but the implementing symbols (`layout.Anchor`, `PageResult.Anchors`, `LinkDestName`) are **absent from the tree** — the entry outlived the code (reverted or never merged).

## Fix

Re-implements the `id → named destination` path by mirroring the existing `bookmarkAnchor` mechanism:

`id` attr → wrap element in a `layout` anchor marker → marker tags its first `PlacedBlock.AnchorName` → renderer surfaces `PageResult.Anchors` → the document layer registers them via `AddNamedDest` during layout → the existing `/Dests` writer + annotation resolver emit the destination.

- **`layout/anchor.go`** (new) — `NewAnchor(inner, name)`, a thin pass-through wrapper modeled on `bookmark.go`; delegates `Measurable` so flex / table-cell shrink-to-fit is unaffected.
- **`layout`** — `PlacedBlock.AnchorName`, `PageResult.Anchors` (`AnchorInfo{Name, Y}`), recorded in `drawBlockNested` at the block's top `y`.
- **`html/converter_dispatch.go`** — wrap the produced element when the node carries an `id` (block-element dispatch, so headings/divs/sections are covered uniformly).
- **`document/document.go`** — register each anchor as a `NamedDest` (`FitH` at the target `y`). Extracts a shared `buildNamedDestArray` so the annotation resolver honors `FitH`/`Top` (links land on the section, not the page top) and agrees with the `/Dests` writer.

No caller `AddNamedDest()` is required — registration is automatic.

## Tests

- `layout/anchor_test.go` — wrapper unit tests (name on first block, overflow not re-tagged, `Measurable` delegation, empty-plan/empty-name safety).
- `html/anchor_dest_test.go` — end-to-end: `html.Convert → WriteTo`, then walks the parsed catalog's `/Dests` to prove the destination is **defined**, asserts `/FitH` and the absence of a dangling `/GoTo`. Verified red before the fix, green after.
- `gofmt -s`, `go vet`, and the full `go test ./... -race` suite pass.

## Scope notes

- Stays on the existing flat `/Dests` dictionary (ISO 32000 §12.3.2.3), which viewers honor. A `/Names → /Dests` name tree is a separate optional enhancement.
- Inline `id`s (e.g. `<span id>` mid-paragraph) flow through the inline path and are out of scope; real anchor targets are block-level.
- The stale `## [0.8.0]` #223 CHANGELOG line is left as released history; this change is logged under `## Unreleased → Fixed`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- **Internal anchor links now register a PDF named destination** — a block-level element with an `id` (`<h2 id="section">`, `<div id="details">`, …) now auto-registers a named destination, so `<a href="#section">` resolves to a direct `/Dest` on the target's page instead of emitting a dangling `/GoTo` string action that jumps nowhere. The HTML walker wraps any block element carrying an `id` in a `layout` anchor marker that tags its first `PlacedBlock`; the renderer surfaces these on `PageResult.Anchors`, and the document layer registers them via `AddNamedDest` during layout — no caller `doc.AddNamedDest()` needed. Resolved links use an `/XYZ` destination at the target's y-position with zoom retained (the same shape auto-bookmarks use, so a link and an outline entry to the same target behave identically); the annotation resolver and the `/Dests` writer share one destination-array builder and agree first-registration-wins on duplicate names. The anchor and bookmark markers now expose an `unwrap` so the layouter still reaches a decorated element's optional interfaces (CSS `clear`, `page-break-inside: avoid`, flex cross-axis stretch) — previously wrapping an element to record its `id` silently disabled those. Inline targets (`<span id>`, the `<a id>`/`<a name>` idiom) are out of scope: the destination is registered only for block-level elements. This restores the `layout.Anchor` behavior documented under 0.8.0 (#223), which was absent from the tree.
+
 ## [0.10.0] - 2026-07-10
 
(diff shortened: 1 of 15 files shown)


kuadrosx:
This PR's content is already on `main` as commit 57e09c6, which landed via #432 (`worktree-consolidated-pr-bugs`). The two are the same patch — `git patch-id --stable` matches exactly (`b9a2c6f6…`) across all 15 files, including the CHANGELOG entry.

```
$ git show b408cca | git patch-id --stable   # this PR
b9a2c6f674bda08bf63de7dcb2c9ebf078b4482c b408cca…
$ git show 57e09c6 | git patch-id --stable   # on main
b9a2c6f674bda08bf63de7dcb2c9ebf078b4482c 57e09c6…
```

Since it was cherry-picked rather than merged, GitHub can't detect it and still shows the branch as conflicting. Nothing here is outstanding — closing as already-landed.

kuadrosx:
Closing as already-landed: the content is on `main` as 57e09c6 (identical patch-id, see above).

PR #423, opened 2026-07-12
layout: fragment tall content across pages and keep it within the page margins

kuadrosx:
## Problem

Two related classes of pagination defect in the fragmentation path.

**A. Content taller than a page was silently clipped and dropped** instead of continuing on the next page:

1. **`Div.PlanLayout`** — the in-flow child loop's `LayoutFull` arm placed a child without checking it fit the remaining height. An unsplittable child (e.g. a fixed-height box) landing across the page boundary was placed anyway, pushing the offset past the page bottom, and every later child followed; the tail was dropped.
2. **`planRow`** — a flex line only broke to the next page once a prior line had already fit (`fittedLineCount > 0`), so a single/first/only line taller than the page was laid out whole and each column's overflow discarded.
3. **`planRow`** again — a flex item can report **`LayoutNothing`**: it placed nothing at all, so it hands back neither blocks nor overflow and its content lives on only in the element itself. `planRow` had no arm for that status. It read the (empty) blocks, laid the line out with the remaining columns and reported `LayoutFull`, so that column's entire content was dropped. The output stayed a well-formed PDF — it paginated, every page looked complete, nothing appeared cut — and the page count moved the *wrong way*: content that was dropped never asks for pages of its own, so the broken render is **shorter** than the correct one and a page-count check reads the loss as an improvement. The shape that exposes it is a flex row used as a row wrapper (a short fixed-width marker plus a tall block column) straddling a page boundary; the status itself is easy to reach there, from either of the two `LayoutNothing` guards below (item 5, and the `Div` zero-progress guard). With `align-items` other than `stretch` the whole column vanished; under the default `stretch` the cross-axis re-layout pass masked it only when the marker happened to be taller than the column's first unbreakable unit.

**B. When content *did* fragment across a page boundary, it overran the page bottom margin** instead of stopping at `pageHeight − topMargin − bottomMargin` and continuing on the next page:

4. A fragmenting **`Div`** under-reported its height: the `LayoutPartial` arm placed the fitted portion but never advanced the running offset (so `Consumed` came back ≈ 0 while a full page of content had been placed), and `innerHeight` omitted the box's own top/bottom spacing. The renderer then packed the following block on top and the total spilled past the bottom margin.
5. A **`<table>`** beginning near the page bottom placed its header rows *and* first body row unconditionally — the split check was gated on `i > headerRowCount`, so it never fired for the first body row — overrunning the margin (and able to strand a header at the very bottom).
6. **`break-inside: avoid` was silently dead on a decorated block.** `Div.PlanLayout`'s keep-together check type-asserted the interface on the child element directly, but this package already wraps elements on the way out of the converter — an element carrying an HTML `id` gets an anchor marker (`anchor.go`), one carrying `bookmark-level` a bookmark anchor (`bookmark.go`) — and neither wrapper reproduces `KeepTogether`. Every decorated block therefore missed the check and was fragmented after all, stranding its leading content in the page's bottom margin while the rest moved on.

A browser fragments all of these across pages, keeps the content, and never draws into the bottom margin.

## Fix

**Fragmentation (content loss)**
- **`Div`** breaks between children when an unsplittable child overflows the remaining height, carrying the remainder to overflow. Guarded on having already placed a child so a lone child taller than a whole page is still drawn (clipping) rather than looping — mirroring the flex `planColumn` guard. Scoped to auto-height, non-clipping containers (`paginateOverflow`), so definite/limited-height, `aspect-ratio`, or `overflow: hidden` boxes keep contain-and-clip semantics.
- **`planRow`** fragments a line in place when its items report `LayoutPartial`: fitted portions stay on the page; a continuation `Flex` pins each unfinished column's overflow to the width it occupied and replaces finished columns with same-width spacers so positions are preserved. A line that does not fit is deferred **whole** to the next page (takes precedence over in-place fragmentation, so a later line never straddles the bottom); an atomic line taller than the page still falls through and is placed, so pagination cannot loop.
- **`planRow`** handles an item that placed nothing: the line is deferred **whole** — to the next page when earlier lines already fit, or, when nothing has been placed yet, by reporting `LayoutNothing` so the renderer relocates the container to a fresh page where the column can make progress. At page top the renderer force-places a `LayoutNothing` element with an effectively unbounded height, so this cannot loop. Mirrors `planColumn`'s `fittedCount == 0` handling and the `Div` zero-progress guard, and is checked *before* the fits/atomic fall-through, which would otherwise place the line and drop the column. Scoped to `paginateOverflow`, so a definite-height flex keeps contain-and-clip semantics.

**Bottom margin**
- **`Div`**: advance the running offset past a fitted partial child so the container reports its true consumed height, and subtract the box's own `spaceBefore`/`spaceAfter` from the space offered to children (mirroring the flex container). A `Div` that places nothing reports `LayoutNothing` (instead of a zero-progress partial), so the renderer relocates it to a fresh page and, at page top, force-places it — preventing an infinite loop for a wrapped child taller than a whole page.
- **`<table>`**: fit-check the first body row too, and when neither header + first row fits, defer the *whole* table (`LayoutNothing`) so the header and its first row move to the next page together — rather than stranding the header. The renderer relocates it and force-places a group taller than a whole page, so pagination still terminates.
- **`Div` keep-together**: assert the interface through `baseElement` so the anchor/bookmark decorators are unwrapped, mirroring the renderer's top-level check in `render_plans.go`. `break-inside: avoid` now behaves identically with and without an `id` or `bookmark-level`.

## Tests

- `layout/div_fragment_test.go`, `layout/flex_fragment_test.go`, `html/flex_fragment_test.go` — content preservation across the overflow chain, lone-oversized-child / atomic-line fallbacks, whole-line deferral (no straddling), the definite-height carve-out, and end-to-end multi-page spanning (HTML → PDF → page count).
- `layout/fragment_page_margin_test.go` — fragmented Div/flex content stays within `[bottomMargin, pageHeight − topMargin]` across pages, and a fragmenting Div reports `Consumed` equal to the content it placed.
- `layout/table_page_margin_test.go` — a table starting near the page bottom renders no row or header below the bottom-margin line, and never emits a header without room for its first body row; plus an end-to-end oversized-group termination guard.

- `layout/flex_fragment_test.go`, `html/flex_fragment_test.go` (content-loss coverage) — a row wrapper whose column reports `LayoutNothing` must defer rather than drop it; the whole page chain (including the renderer's `LayoutNothing` relocation and page-top force-place) must preserve every leaf block; a later line of a wrapped container must be deferred whole. End-to-end, the same row wrapper is rendered at every offset from 600pt to 780pt of leading fill — the window that loses content is only a few points wide, so a single fixed offset sails past the bug — and the assertion is the set of text tokens that reach the PDF, with the same content in a plain **block** wrapper as a passing control. These deliberately do not assert on page count: it moves the wrong way when content is dropped.
- `html/break_inside_avoid_test.go` — a keep-together block carrying an `id` or `bookmark-level`, nested in a wrapper, must paginate identically to an undecorated one (per-page word counts compared) and must not draw into the bottom margin.

New tests fail before the fix and pass after. Full `go test ./...`, `gofmt -s`, and `go vet` are clean.

## Also in this PR: a CI lint fix

The ubuntu job started failing at the **Lint** step — before Build and Test run — on six staticcheck findings in `font/subset_test.go`, `reader/flatten.go` and `sign/signer.go`, none of which this branch touches. Cause: `ci.yml` uses `golangci-lint-action` with `version: latest` and the repo carries no golangci config, so the toolchain floats; v2.13.1 (released 2026-08-20) enables checks v2.11.x did not. The same six findings reproduce on an unmodified `origin/main` with that version, so `main` is red for the same reason.

Fixed here so this PR can go green: the two deprecated-API sites (SA1019) move to `NewPdfIndirectReference` / `NewPdfArray` — the `flatten.go` one was also inconsistent with the sibling `Annots` rebuild 50 lines above — and `ParsePKCS12`'s always-true error check (SA4023, from the `decodePKCS12` placeholder that always errors) is suppressed in the file that owns the placeholder rather than removed; collapsing it would push the identical finding onto every caller (`export/cabi_v062.go:371`). Verified with the exact version CI installs: 0 issues.

Worth considering separately: pinning `version:` in `ci.yml` so a lint release cannot turn unrelated PRs red.


kuadrosx:
Rebased onto v0.10.1 — the conflict with `main` is resolved and this is mergeable again. See the merge commit message for the reasoning behind each resolution.

On the red CI, neither job is caused by this PR:

- **`test (ubuntu-latest)`** fails only on the 3 pre-existing staticcheck SA1019 errors in `document/anchor_dest_offset_test.go`. These fail on `main` too, and because the lint job runs whole-tree with `only-new-issues: false` they fail every open PR. #436 fixes them (lint goes to `0 issues`). This PR's own lint issues are fixed.
- **`test (windows-latest)`** fails on exactly the same 31 tests as `main` — golden-file comparisons plus the new qpdf AES-256 fixtures. Diffed against the `main` run: 0 new failures, 0 fixed.

`go test ./...`, `gofmt -s -l .` and `golangci-lint run` are all clean locally apart from those pre-existing items.

kuadrosx:
Updated onto `main` with #436 merged. **All four checks are green** — `test (ubuntu-latest)`, `test (windows-latest)`, `test (macos-latest)`, `fuzz` — so the red CI noted earlier is gone and this is ready for review on its own merits.

Clean merge, no conflicts. Locally on the merged branch: `go test ./...` passes, `gofmt -s -l .` clean, `golangci-lint run` → `0 issues`.

kuadrosx:
@carlos7ags man can you help me reviewing the PRs? :'(

carlos7ags (dismissed):

kuadrosx:
@carlos7ags 🙏🏻🙏🏻🙏🏻

carlos7ags (approved):

kuadrosx:
Hmm, why didn't it merge?@carlos7ags

PR #424, opened 2026-07-12
html, layout: font-metric line-height:normal, correct length line-height, and white-space:nowrap

kuadrosx:
## Problem

Several related line-box defects, all visible against headless Chrome rendering the same HTML/CSS:

1. **`line-height: normal`** resolved to a flat `fontSize * 1.2`, ignoring the font's own vertical metrics. Faces that declare a large line-gap (e.g. Poppins) rendered visibly more compact than every browser, drifting pagination over long documents.
2. **`line-height: <length>`** (e.g. `line-height: 20px`) was eagerly divided by the *current* font-size when its declaration was applied, so declaration order mattered: `line-height:20px; font-size:11px` divided by the not-yet-applied default and produced a too-short line box, making a single glyph ride high instead of centering (breaks the `line-height == height` vertical-centering idiom).
3. **`white-space: nowrap`** was ignored by word-wrapping: an over-long unbreakable token was character-broken across lines instead of overflowing on one line. It was also dropped when set on an inline element (it inherits) rather than the block, and defeated by collapsible whitespace around such an element.

## Fix

- **normal leading from metrics**: add `Face.LineGap()` (mirroring the `Ascent`/`Descent` `USE_TYPO_METRICS` selection) and `EmbeddedFont.NormalLineHeight()`; a shared `resolveLineHeight` takes the max normal leading over all runs on the line (so mixed-font lines size to the tallest inline box), falling back to `fontSize*1.2` for base-14 fonts. Only the CSS path defaults to a `LeadingNormal` sentinel; direct API callers are unchanged.
- **length line-height**: re-resolve a standalone `line-height` against the element's *final* font-size after all its declarations are applied, so order no longer matters. Half-leading then centers the single line correctly.
- **nowrap**: a `Paragraph.noWrap` flag, threaded from CSS (recorded per run since `white-space` inherits, and applied when all of a paragraph's text runs come from a nowrap context). It suppresses both long-word character-breaking and space-boundary wrapping in both word-wrap implementations, takes precedence over `word-break: break-all`, is preserved across page/column-split continuations, and drives ellipsis truncation of the single overflowing line. `MinWidth` returns the full single-line width under nowrap.

## Tests

Unit and HTML-level regression tests for each: `NormalLineHeight`/`MeasureString` across six font families (TrueType and CFF outlines); half-leading centering of a single line in a taller box; length line-height resolving identically in either property order; nowrap keeping over-long tokens, multi-word runs, inline-element nowrap (incl. pretty-printed whitespace, nested inline, links) on one line; nowrap+ellipsis; and the documented mixed-content limitation. Cross-checked against headless Chrome across all six fonts.


kuadrosx:
Rebased onto v0.10.1 — the conflict with `main` is resolved and this is mergeable again. See the merge commit message for the reasoning behind each resolution.

On the red CI, neither job is caused by this PR:

- **`test (ubuntu-latest)`** fails only on the 3 pre-existing staticcheck SA1019 errors in `document/anchor_dest_offset_test.go`. These fail on `main` too, and because the lint job runs whole-tree with `only-new-issues: false` they fail every open PR. #436 fixes them (lint goes to `0 issues`). This PR's own lint issues are fixed.
- **`test (windows-latest)`** fails on exactly the same 31 tests as `main` — golden-file comparisons plus the new qpdf AES-256 fixtures. Diffed against the `main` run: 0 new failures, 0 fixed.

`go test ./...`, `gofmt -s -l .` and `golangci-lint run` are all clean locally apart from those pre-existing items.

kuadrosx:
Updated onto `main` with #436 merged. **All four checks are green** — `test (ubuntu-latest)`, `test (windows-latest)`, `test (macos-latest)`, `fuzz` — so the red CI noted earlier is gone and this is ready for review on its own merits.

Clean merge, no conflicts. Locally on the merged branch: `go test ./...` passes, `gofmt -s -l .` clean, `golangci-lint run` → `0 issues`.

kuadrosx:
@carlos7ags Let me know if you need anything for this one.

kuadrosx:
@carlos7ags 🙏

PR #436, opened 2026-08-14
Fix CI on main: staticcheck lint and the Windows job

kuadrosx:
`main` is currently red: the `golangci-lint` step added in #434 reports three staticcheck SA1019 errors, all in `document/anchor_dest_offset_test.go`, which reads the deprecated `PdfIndirectReference.ObjectNumber` field directly.

```
document/anchor_dest_offset_test.go:64:34:  SA1019: pageRef.ObjectNumber is deprecated … Use [PdfIndirectReference.Num]
document/anchor_dest_offset_test.go:111:34: SA1019: pageRef.ObjectNumber is deprecated … Use [PdfIndirectReference.Num]
document/anchor_dest_offset_test.go:133:64: SA1019: ref.ObjectNumber is deprecated … Use [PdfIndirectReference.Num]
```

This switches the three reads to the documented replacement, `Num()`. Test-only change, no behavior difference — `Num()` is `func (r *PdfIndirectReference) Num() int { return r.ObjectNumber }`.

With this applied, `golangci-lint run` reports `0 issues` for the whole tree, and `go test ./document/` still passes.

Worth flagging: because the lint job runs over the whole tree with `only-new-issues: false`, these three errors currently fail the lint job on **every open PR** as well, not just on `main` — so this unblocks them too.

(Separately, the `test (windows-latest)` job is also failing on `main`, on golden-file comparisons and the new qpdf AES-256 fixtures. That looks like a line-ending/platform issue and is untouched here.)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,29 @@
+# Normalize line endings so the test suite passes on Windows.
+#
+# Without this, Git for Windows (and the windows-latest CI runner, which sets
+# core.autocrlf=true) rewrites LF to CRLF on checkout. Two things break:
+#
+#   1. Byte-exact comparisons against checked-in fixtures — the CSS support
+#      doc (TestCSSDocsInSync) and the style/paragraph .golden corpora — see a
+#      stray \r on every line and report every case as stale.
+#   2. Binary fixtures that happen to contain no NUL byte in their first 8000
+#      bytes, which is all Git inspects to guess text-vs-binary. Several of the
+#      small qpdf-generated encrypted PDFs land on the wrong side of that
+#      heuristic, so CRLF expansion corrupts their encrypted streams and
+#      decryption fails with "ciphertext length N is not block-aligned".
+#
+# Checking text in as LF everywhere and pinning the working-tree ending to LF
+# keeps both honest on every platform.
+* text=auto eol=lf
+
+# Fixtures that must survive byte-for-byte. Being explicit matters on the *add*
+# side too: under `text=auto` alone, a binary file Git mistakes for text has its
+# CRLF runs normalized to LF in the index, i.e. corrupted at commit time. This
+# also stops Git from trying to diff or merge them.
+#
+# .afm is deliberately absent — Adobe Font Metrics really is a text format, so
+# it is left diffable and covered by the eol=lf rule above.
+*.pdf  binary
+*.png  binary
+*.ttf  binary
+*.otf  binary
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,6 +133,13 @@ jobs:
           rm -rf "$work"
 
       - name: ZUGFeRD XSD validation
+        # This block is bash, and on windows-latest the default shell is
+        # PowerShell, which rejects it with "Missing '(' after 'if'". The step
+        # is deliberately not Linux-only — the schema dir is wired up only on
+        # Linux, while the test itself runs everywhere and skips when
+        # FOLIO_FACTURX_XSD_DIR is unset — so pin the shell instead of guarding
+        # the step.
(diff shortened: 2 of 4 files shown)


carlos7ags (dismissed):

kuadrosx:
hmm the CI is red but it looks unrelated to this PR

kuadrosx:
Expanded to cover the other red job as well, so `main` goes green on all four checks.

`test (windows-latest)` fails on `main` with **31** failing tests, from two causes:

**30 are CRLF corruption.** The runner uses Git for Windows with `core.autocrlf=true`, and the repo has no `.gitattributes`, so LF becomes CRLF on checkout:

- Byte-exact fixture comparisons see a stray `\r` per line — `TestCSSDocsInSync` reports `docs/CSS_SUPPORT.md` stale (`on disk: "# Folio CSS support\r"`), and every `TestStyleGolden` / `TestParagraphGolden` case mismatches.
- Git only checks the first 8000 bytes for a NUL when guessing text-vs-binary, and several of the small qpdf encrypted PDFs added in #434 have none, so they get expanded as text. That corrupts their encrypted streams — `aes256.pdf` grows 1425 → 1461 bytes (one byte per LF) and `TestParseQpdfEncrypted` fails with `ciphertext length 81 is not block-aligned`.

**The 31st is a real portability bug.** `TestFallbackFontAbsoluteSkipsBaseFS` built its "absolute" path as a leading separator plus a relative path. On Windows `\does\not\exist\x.ttf` is rooted but *not* absolute (`filepath.IsAbs` wants a volume like `C:\`), so `loadFallbackFont` treated it as relative and consulted `BaseFS` — precisely what the test asserts must not happen. Now uses `filepath.Abs`.

No file contents change: the stored blobs are already LF, so `.gitattributes` only affects how they are checked out. The explicit `binary` markers also matter on the *add* side, where `text=auto` alone would normalize CRLF runs inside a misdetected binary and corrupt it at commit time. `.afm` is deliberately left out — it genuinely is a text format.

**How this was verified without a Windows machine:** cloning with `-c core.autocrlf=true` reproduces the runner's checkout. Before the fix, 30 of the 31 failures reproduce on macOS with byte-identical error messages; after, the suite is green. On this branch: `go test ./...` clean, `gofmt -s -l .` clean, `golangci-lint run` → `0 issues`.

kuadrosx:
All four checks green now: `test (ubuntu-latest)`, `test (windows-latest)`, `test (macos-latest)`, `fuzz`.

Fixing the 31 test failures uncovered one more problem hiding behind them. The Windows job then got as far as **ZUGFeRD XSD validation** and died there:

```
ParserError: Missing '(' after 'if' in if statement.
```

That step's script is bash, but `windows-latest` defaults to PowerShell. It had simply never been reached before, because the `Test` step failed first and later steps are skipped — so the breakage was masked by the CRLF failures.

Fixed with `shell: bash` rather than a Linux guard, to preserve the step's intent: the schema directory is wired up only on Linux, while the test itself runs on every platform and skips cleanly when `FOLIO_FACTURX_XSD_DIR` is unset. I checked the rest of the workflow — it is the only step with a shell-script body that is neither Linux-guarded nor given an explicit `shell:`, and it is the last non-Linux-guarded step in the job, which is why the job now runs to completion.

Summary of the three commits:
| commit | fixes |
|---|---|
| `7c5496a` | 3 staticcheck SA1019 errors → lint job (`0 issues` whole-tree) |
| `7a9ee0c` | `.gitattributes` + a non-portable absolute path → 31 Windows test failures |
| `603789d` | bash step running under PowerShell → Windows job completes |

kuadrosx:
@carlos7ags claude went ahead and fix the issues, can you help me with the merge? so the other PRs can get merge

